PR #16, opened 2018-02-17
Update workbench

zhirzh:
Updates the entire codebase from react-scripts v0.6.1 to the current latest v1.1.1


package.json
```diff
@@ -3,53 +3,60 @@
   "version": "0.1.0",
   "homepage": "http://vega.github.io/editor",
   "devDependencies": {
-    "autoprefixer": "6.4.1",
+    "autoprefixer": "7.1.6",
```
domoritz: Do you think we can remove a lot of these because we use `react-scripts`?

Here is what a clean app looks like

```json
{
  "name": "my-app",
  "version": "0.1.0",
  "private": true,
  "dependencies": {
    "react": "^16.2.0",
    "react-dom": "^16.2.0",
    "react-scripts": "1.1.1"
  },
  "scripts": {
    "start": "react-scripts start",
    "build": "react-scripts build",
    "test": "react-scripts test --env=jsdom",
    "eject": "react-scripts eject"
  }
}
```

zhirzh: AFAIK, we can't do that. We need to eject react-scripts to modify webpack configs and other scripts. Once ejected, react-scripts dumps all its packages, scripts and package settings into `package.json`

Issue #12, opened 2018-01-19
Combine the editor instances


domoritz:
Right now we create new monaco instances when we open the vega spec in vl mode. This prevents undo/redo.

domoritz:
Just to explain this a bit more. The problem is that react recreates components when a subtree of the dom changes. This means we recreate the monaco editor instance in some case and erase the (undo) history.

domoritz:
We have fixed the issue when the Vega editor instance is opened side by side with Vega-Lite. However, switching from Vega to Vega-Lite and vice versa still creates a new editor.

algomaster99:
@domoritz On switching from vega to vega-lite, we can just change the editorString right? This is how this issue can be solved right?

domoritz:
Yeah, maybe. The issue is that react will re-create the component instead of reusing the editor we already have. This is one of the most important issues we have right now. I would love to support undo across different modes for example.

algomaster99:
Oh okay. You mean undo can be done even if we switch editors?

domoritz:
It can't right now because we create a new Monaco instance.

algomaster99:
Yeah right. I'll try to fix this issue.
What I am thinking is that we just change editorStrings. This will prevent from creating a new instance and also undoing will be possible. I'll look into it and let you know the progress soon

domoritz:
Great. It's a bit tricky to do but shouldn't be impossible.

algomaster99:
I have unassigned this issue to myself so others can also take a look.

domoritz:
We support undo/redo across modes now.

PR #21, opened 2018-02-18
Prevent creation of new monaco instances

saurabh0402:
Solves #12


domoritz:
I get an error when I launch the editor

screen shot 2018-02-17 at 22 37 49

PR #22, opened 2018-02-19
Close portal on esc and outside click

saurabh0402:


domoritz:
I think #23 is a cleaner approach to this issue. Let me know if you disagree.

PR #27, opened 2018-02-20
render one <Route /> per URL

### LINKED ISSUE #25: tab creates two app containers
Press tab while not focused on the editor:

editor

zhirzh:
Fixes #25.

### DIFF
--- a/src/components/app-shell.tsx
+++ b/src/components/app-shell.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {Route} from 'react-router-dom';
+import {Route, Switch} from 'react-router-dom';
 
 import App from './app';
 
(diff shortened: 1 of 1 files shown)


src/components/app-shell.tsx
```diff
@@ -15,11 +15,13 @@ export default class AppShell extends React.Component<Props> {
   public render() {
     return (
       <div>
-        <Route path='/' component={App} />
-        <Route path='/edited' component={App} />
-        <Route path='/gist/:mode/:username/:id' component={App} />
-        <Route path='/examples/:mode/:example_name' component={App} />
-        <Route path='/custom/:mode' component={App} />
+        <Switch>
+          <Route path='/' exact component={App} />
```
domoritz: Is exact still needed here now that we use switch?

zhirzh: Without `exact`, all paths will always mount the first `<Route />` since every path begins with `/`.

PR #28, opened 2018-02-20
Add validation to gist form

saurabh0402:


domoritz:
I think we should accept both URLs and user/gist-id strings. Can you change the matching to include the latter?

saurabh0402:
I have changed the Regular expression to allow user/gist-id.

domoritz:
Thanks. Merged with some code fixes.

PR #32, opened 2018-02-21
Fix wrong url when user/gist-id used to load gists

saurabh0402:


domoritz:
Please fix oint issues first. 

Also, isn't there a simpler way to write this?

PR #34, opened 2018-02-22
Redesign gist loading

### LINKED ISSUE #11: Export to gists

### LINKED ISSUE #31: Clean up load from gist
* Don't rely on a service.
* Clean up the validation. The regexes are too complicated and not accurate (suggested by @zhirzh).

zhirzh:
fixes #31
fixes #11


domoritz:
I think we can make the experience a bit better. Here are a few suggestions
* Provide examples for both the URL and the filename
* Have a default filename (e.g. `spec.vl.json` for VL and `spec.vg.json` for Vega)
  * For this we might want to move the Vega/Vega-Lite picker to the top. It should probably become a choice instead of a button. 
* When you already have a gist loaded, we should autofill the URL and filename fields with the current values. 

Let me know what you think.

zhirzh:
> When you already have a gist loaded, we should autofill the URL and filename fields with the current values.

Can you please explain what you mean by this?

domoritz:
Btw, I think it's a bit odd to have the URL and the filename separate. A URL is a "Uniform Resource Locator" and thus should point to a specific file.

zhirzh:
What about adding flexibility for the user to enter gist URL? What forms of input should we allow?

AFAIK, a gist can be uniquely identified by a number of patterns which are represented by [this regexp](https://regexper.com/#^(?:(?:protocol)?(?:domain))?(?:user)?(?:gist-id)(?:revision)?$) and a more concrete regexp [here](https://regexper.com/#^(?:(?:https?:\/\/)?(?:gist\.github\.com))?((?=.{1,39})(?:[0-9a-z]+(?:-[0-9a-z]+)*))?([0-9a-z]{32}|[0-9a-z]{8})([0-9a-z]{32})?$).

domoritz:
I think we want the example to be copyable (or click + insert) and not just in the background. By click+insert, I mean a text that you can click on and it automatically fills the text input. This can be useful for someone who just wants to try gist loading to see how it works. 

Also, I’d still put the choice for Vega + Vega-Lite to the top and then just have a “Load” button.

zhirzh:
I can add the choice for Vega + Vega-Lite to the top here but it seems that the example thing needs some more thought and its own space.

zhirzh:
I have added the gist type switcher but the design is [super basic](https://github.com/zhirzh/editor/blob/f4716b7670e1618d48bccb964aa9faf7968c8a36/src/components/header/index.css#L224-L230). I think it'd be best if all the functionality is first made stable and then we can focus on design.

domoritz:
Sounds good. Let's add the example links I talked about and then merge.

domoritz:
Thank you!

Issue #5, opened 2017-12-03
Export Button


domoritz:
From https://github.com/vega/vega-editor/issues/184

domoritz:
The export button/print mode should just show the plot so you can print it.

ishansaksena:
https://github.com/vega/vega-embed

domoritz:
I don't think so. When you view a chart in full screen you can also save or print as PDF.

saurabh0402:
@domoritz If it's fine with you I would like to work on this.

domoritz:
I know that @ishansaksena already started working on this. He might have some starter code.

ydlamba:
Working on export button.

domoritz:
🎉

domoritz:
Do you have a suggestion where to put the export button? I feel out toolbar in the bottom right is overfowing. Also, we have a parse button above the editor in manual mode. Later, we also want to place the format button (https://github.com/vega/editor/issues/37) somewhere.

ydlamba:
I think all actionable buttons should be together. So that we can refer them as a toolbar.
![image](https://user-images.githubusercontent.com/21237466/37012644-bd0d7558-211b-11e8-8a43-2b6abd70ff79.png)

domoritz:
As I said, another option is right above the editor. I think where you have it is good for now.

PR #42, opened 2018-03-06
Adding: Export Button

ydlamba:
**Adding: Export Button**

- Using export button one can export the chart as `png` or `svg` depending upon the renderer.

#5.


src/components/renderer/renderer.tsx
```diff
@@ -44,6 +45,20 @@ export default class Editor extends React.Component<Props> {
...
     }
 
+    if (this.props.export) {
+      const ext = this.props.renderer === 'canvas' ? 'png' : 'svg';
+      const url = view.toImageURL(ext);
+      url.then(function(url) {
+        var link = document.createElement('a');
+        link.setAttribute('href', url);
+        link.setAttribute('target', '_blank');
+        link.setAttribute('download', 'export.'+ ext);
```
domoritz: instead of downloading, can we just open it in a new tab?

src/components/renderer/renderer.tsx
```diff
@@ -44,6 +45,20 @@ export default class Editor extends React.Component<Props> {
       }
     }
 
+    if (this.props.export) {
+      const ext = this.props.renderer === 'canvas' ? 'png' : 'svg';
+      const url = view.toImageURL(ext);
+      url.then(function(url) {
```
domoritz: => instead of function

domoritz:
This is a great start. Can you try an interactive example and make sure that the exported visualization is the of the current state, not the initial state?

src/components/renderer/renderer.tsx
ydlamba: Actually, View API gives URI for `png` whose support is removed by many browsers in a new tab, they did it for security reasons. Now we have the only option of downloading it. Still, we can open SVG in a new tab.

ydlamba:
Currently, the current state (after interaction)  is not stored that's why it is exporting the initial state.

ydlamba:
To fix this we need to set a new state that preserves the user interactions. But I am not getting which vega module is doing the interaction part.

domoritz:
What happens if you use the existing Vega view instead of creating a new one? Does that contain the interactive state?

ydlamba:
I am already using the existing one but it gets reset each time as component updates. Now I am trying by storing view as a state.

src/components/renderer/renderer.tsx
domoritz: I think for SVG it's more important. My use case is that I want to print the SVG as PDF. It's not important for PNGs.

ydlamba:
@domoritz Using the existing view is perfectly exporting with interactive states so setting view as a state doesn't make a point plus requested changes are done. You can have a look.

PR #46, opened 2018-03-09
Fix: Remove unnecessary scrollbar (in few browsers)

ydlamba:
**Fix: Remove extra scroll (in few browsers)**

- In few browsers (like Chrome 64.0), there is an unnecessary disabled scrollbar associated with chart container, which looks like this:
image

- After fix:
image

### DIFF
--- a/src/components/viz-pane/index.css
+++ b/src/components/viz-pane/index.css
@@ -5,3 +5,7 @@
   padding: 10px;
   overflow: auto;
 }
+
+.Pane::-webkit-scrollbar {
+    display: none;
+}


domoritz:
Does the scrollbar also disappear if we set overflow to none? I'm not quite sure why it's necessary to hide a scrollbar (and I'm a bit scared to do so).

ydlamba:
Well, there are two scrollbars right now in which the outer one is **always** disabled and the inner one is perfectly doing its job. So in my opinion, we don't need an extra scrollbar for any purpose. 
Plus `display: none` is a better way to remove it than `overflow: hidden` since the former removes the element from the page while making overflow hidden leave its space behind.

`display: none`:
image

`overflow: hidden`:
image

Over to you.

PR #47, opened 2018-03-10
[WIP] Use vega-loader for validating links

### LINKED ISSUE #30: Set baseurl based on post message
See https://github.com/vega/vega-embed/issues/55

Basically, when we get a spec via post message, we should use try to resolve URLs as a) absolute, b) relative to the editor, c) relative to the caller. This can be implemented by creating a custom vega loader.

ydlamba:
**Use vega-loader for validating links**

- It first checks whether the link is `absolute` or `relative to the editor`. If gets `404` then check whether it is `relative to the caller` and if still gets an error then throws it. 

Fixes #30.

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -25,6 +25,8 @@ class App extends React.Component<Props & {match: any, location: any}> {
         if (!data.spec) {
           return;
         }
+        // setting baseURL as event's origin
+        this.props.setBaseUrl(evt.origin);
         console.info('[Vega-Editor] Received Message', evt.origin, data);
         // send acknowledgement
         const parsed = JSON.parse(data.spec);
@@ -141,6 +143,9 @@ const mapDispatchToProps = function(dispatch) {
     },
     updateVegaLiteSpec: (val) => {
       dispatch(EditorActions.updateVegaLiteSpec(val));
+    },
+    setBaseUrl: (val) => {
+      dispatch(EditorActions.setBaseUrl(val));
     },
     setVegaExample: (example: string, val) => {
       dispatch(EditorActions.setVegaExample(example, val));
--- a/src/components/renderer/renderer.tsx
+++ b/src/components/renderer/renderer.tsx
@@ -14,6 +14,7 @@ type Props = {
   mode?: string;
   tooltip?: boolean;
   export?: boolean;
+  baseURL?: string;
 };
 
 export default class Editor extends React.Component<Props> {
@@ -26,10 +27,22 @@ export default class Editor extends React.Component<Props> {
 
     const runtime = vega.parse(props.vegaSpec);
 
-    Editor.view = new vega.View(runtime)
-      .logLevel(vega.Warn)
-      .initialize(Editor.chart);
+    const loader = vega.loader();
+    const originalLoad = loader.load.bind(loader);
+    // Custom Loader
+    loader.load = async(url, options) => {
+      try {
+        return await originalLoad(url, {options, ...{baseURL: this.props.baseURL}});
+      } catch {
+        return await originalLoad(url, options);
+      }
+    };
 
+    Editor.view = new vega.View(runtime, {
+      loader: loader,
+      logLevel: vega.Warn,
+    })
+    .initialize(Editor.chart);
   }
   public renderVega(props) {
     Editor.view.renderer(props.renderer);
@@ -69,7 +82,7 @@ export default class Editor extends React.Component<Props> {
     this.renderVega(this.props);
   }
   public componentDidUpdate(prevProps) {
-    if (prevProps.vegaSpec !== this.props.vegaSpec || prevProps.vegaLiteSpec !== this.props.vegaLiteSpec) this.initilizeView(this.props);
+    if (prevProps.vegaSpec !== this.props.vegaSpec || prevProps.vegaLiteSpec !== this.props.vegaLiteSpec || prevProps.baseURL !== this.props.baseURL) this.initilizeView(this.props);
     this.renderVega(this.props);
   }
   public render() {
--- a/src/components/renderer/renderer.tsx
+++ b/src/components/renderer/renderer.tsx
@@ -32,7 +32,8 @@ export default class Editor extends React.Component<Props> {
     // Custom Loader
     loader.load = async(url, options) => {
       try {
-        return await originalLoad(url, {options, ...{baseURL: this.props.baseURL}});
+        if (options) return await originalLoad(url, {...options, ...{baseURL: this.props.baseURL}});
+        return await originalLoad(url, {baseURL: this.props.baseURL});
       } catch {
         return await originalLoad(url, options);
       }
--- a/src/actions/editor.ts
+++ b/src/actions/editor.ts
@@ -3,6 +3,7 @@ import {Mode, Renderer} from '../constants';
 export const EXPORT_VEGA: 'EXPORT_VEGA' = 'EXPORT_VEGA';
 export const LOG_ERROR: 'LOG_ERROR' = 'LOG_ERROR';
 export const PARSE_SPEC: 'PARSE_SPEC' = 'PARSE_SPEC';
+export const SET_BASEURL: 'SET_BASEURL' = 'SET_BASEURL';
 export const SET_GIST_VEGA_LITE_SPEC: 'SET_GIST_VEGA_LITE_SPEC' = 'SET_GIST_VEGA_LITE_SPEC';
 export const SET_GIST_VEGA_SPEC: 'SET_GIST_VEGA_SPEC' = 'SET_GIST_VEGA_SPEC';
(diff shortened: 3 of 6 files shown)


ydlamba:
Currently, it is implemented for dataURL (like `stocks.csv`, etc). Should I implement the same for schema link?

src/components/app.tsx
```diff
@@ -17,35 +17,68 @@ import VizPane from './viz-pane';
...
+    }
+    App.evt = evt;
+    App.data = evt.data;
+    console.info('[Vega-Editor] Received Message', evt.origin, App.data);
+    // send acknowledgment
+    const parsed = JSON.parse(App.data.spec);
+    const dataURL = parsed['data']['url'];
+
+    // Checking wheather link is absolute or relative to editor
+    this.loaderPromise({baseURL: 'https://vega.github.io/editor/'}, dataURL).then((data) => {
```
domoritz: I think we should first check whether the URL is absolute. In that case, we won't do anything.

src/components/app.tsx
```diff
@@ -17,35 +17,68 @@ import VizPane from './viz-pane';
...
+      this.updateVega(parsed);
+    }).catch((error) => {
+      if (error.target.status === 404) {
+        // Relative to caller
+        this.loaderPromise({baseURL: evt.origin}, dataURL).then(() => {
+          parsed['data']['url'] = evt.origin + dataURL;
```
domoritz: This won't work if we have a nested spec. Instead of this approach, we should set a baseUrl in the state and create a custom Vega loader. Here is some inspiration: https://github.com/domoritz/draco/blob/master/draco-tools/src/shared/js/components/Visualization.js#L50

src/components/app.tsx
ydlamba: `vega-loader` already checking whether it is absolute or not. If not then I am checking it relative to th editor.

domoritz: I think we should first check whether the URL is absolute. In that case, we won't do anything.

domoritz:
> Currently, it is implemented for dataURL (like stocks.csv, etc). Should I implement the same for schema link?

This is irrelevant with the fix I am proposing above.

src/components/app.tsx
ydlamba: `vega-loader` already checks whether it absolute or not. If not then I am checking it with relative to the editor which means we aren't doing anything when URL is absolute. I tested it.

domoritz: Okay, this will might change when we switch to a custom loader.

domoritz:
I'd love to get this in. Can I help with anything @ydlamba?

ydlamba:
I understand your approach, working on it.

src/components/app.tsx
```diff
@@ -25,9 +26,17 @@ class App extends React.Component<Props & {match: any, location: any}> {
         if (!data.spec) {
           return;
         }
+        this.props.setBaseUrl(evt.origin);
         console.info('[Vega-Editor] Received Message', evt.origin, data);
         // send acknowledgement
         const parsed = JSON.parse(data.spec);
+        let dataURL = parsed['data']['url'];
```
domoritz: Let's delete this. We shouldn't expect a spec to have a URL data source.

src/components/app.tsx
```diff
@@ -49,6 +58,23 @@ class App extends React.Component<Props & {match: any, location: any}> {
...
     this.setSpecInUrl(parameter);
   }
+  /**
+   * Return Promise with updated link via vega-loader
+   */
+  public updateLink(url: string): any {
+    const editorURL = 'https://vega.github.io';
+    const loader = vega.loader();
+    const options = {
+      baseURL: this.props.baseURL
```
domoritz: Instead of forcing the baseURL, try to check for every request whether it makes sense to change the baseURL. First, check whether there is data at the `evt.origin` that you grabbed above. If not, use the editor as the base.

src/components/app.tsx
```diff
@@ -49,6 +59,21 @@ class App extends React.Component<Props & {match: any, location: any}> {
     const parameter = this.props.match.params;
     this.setSpecInUrl(parameter);
   }
+  /**
+   * Async function that updates the baseURL (if gets 404)
+   */
+  public async updateBaseUrl(url: string){
```
domoritz: This can be deleted

src/components/app.tsx
```diff
@@ -6,28 +6,38 @@ import * as React from 'react';
...
         const data = evt.data;
         if (!data.spec) {
           return;
         }
+        // Setting BaseURL as event's origin
+        this.props.setBaseUrl(evt.origin);
         console.info('[Vega-Editor] Received Message', evt.origin, data);
         // send acknowledgement
         const parsed = JSON.parse(data.spec);
+        if(parsed['data']['url']) {
```
domoritz: delete this as well

src/components/app.tsx
```diff
@@ -134,6 +159,12 @@ class App extends React.Component<Props & {match: any, location: any}> {
   }
 }
 
+function mapStateToProps(state: State, ownProps) {
```
domoritz: delete this as well

src/components/app.tsx
```diff
@@ -160,4 +194,4 @@ const mapDispatchToProps = function(dispatch) {
   };
 };
 
-export default withRouter(connect(null, mapDispatchToProps)(App));
+export default withRouter(connect(mapStateToProps, mapDispatchToProps)(App));
```
domoritz: you can undo this now

src/components/renderer/renderer.tsx
```diff
@@ -26,10 +27,11 @@ export default class Editor extends React.Component<Props> {
 
     const runtime = vega.parse(props.vegaSpec);
 
-    Editor.view = new vega.View(runtime)
-      .logLevel(vega.Warn)
-      .initialize(Editor.chart);
-
+    Editor.view = new vega.View(runtime, {
```
domoritz: Use something like this instead

```
const runtime = vega.parse(props.vegaSpec);

    const loader = vega.loader();
    const originalHttp = loader.http;
    loader.http = async(url, options) => {
      try {
		return await originalHttp(url, {options, ...{baseURL: this.props.baseURL}});
      } catch {
        return await originalHttp(url, options);
      }
    };

    Editor.view = new vega.View(runtime, {
      loader: loader,
      logLevel: vega.Warn,
    })
    .initialize(Editor.chart);
```

I haven't tested this but I think this communicates the idea.

ydlamba: I think there is some problem with `loader.http`, its baseURL option is not working. Instead `loader.load` is working fine.

ydlamba:
PR is now updated. I have tested this and working as expected (just with few minor changes). You can also have a look.

src/components/renderer/renderer.tsx
```diff
@@ -26,10 +27,22 @@ export default class Editor extends React.Component<Props> {
...
 
-    Editor.view = new vega.View(runtime)
-      .logLevel(vega.Warn)
-      .initialize(Editor.chart);
+    const loader = vega.loader();
+    const originalLoad = loader.load.bind(loader);
+    // Custom Loader
+    loader.load = async(url, options) => {
+      try {
+        return await originalLoad(url, {options, ...{baseURL: this.props.baseURL}});
```
domoritz: 1. it should be ...options, right

domoritz:
Sweet, just one comment.

src/components/renderer/renderer.tsx
ydlamba: I guess you are thinking about options in options. We need to add baseURL directly in the options.
1. `...options` will append `baseURL` in options of options (nested object) while `options` will directly add baseURL in options. I tested it, working fine.
2. Adding an if-else block will work fine.
3. Ignored for now.

ydlamba: So, if we go for `loader.load(url, {options: {baseURL: this.props.baseURL}})` it won't work.

ydlamba:
No changes are done, let me know if I am missing something.

src/components/renderer/renderer.tsx
domoritz: I'm confused. I thought the behavior of ...options and just options is flipped from how you describe it. I will test it when I'm.at a computer again.

ydlamba: Yeah. This is correct. I was thinking it the wrong way, nevermind. I've updated the PR.

domoritz: Last question. Why override load and not http?

ydlamba: Because `loader.http` method performs an HTTP request with a post-sanitization URL i.e we don't have baseURL option with this method. On the other hand, `loader.load` does all processing itself.
For more info check this https://github.com/vega/vega-loader/issues/7.

PR #59, opened 2018-04-14
Apply config to visualization that is posted from vega-embed and upgrade vega-lite

### LINKED ISSUE #55: Update view to use the config which is sent from vega-embed
After https://github.com/vega/vega-embed/pull/66.

ydlamba:
**Apply config to visualization that is posted from vega-embed and upgrade vega-lite**

- Upgrade vega-lite to 2.3.1
- Uses `mergeDeep` to merge config into the spec. 
- Fixes #55.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "redux-thunk": "^2.2.0",
     "vega": "github:vega/vega#v3.2.1",
     "vega-datasets": "^1.18.0",
-    "vega-lite": "github:vega/vega-lite#v2.3.0",
+    "vega-lite": "github:vega/vega-lite#v2.3.1",
     "vega-schema-url-parser": "^1.0.0",
     "vega-tooltip": "^0.7.0"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -74,7 +74,7 @@
     "redux-thunk": "^2.2.0",
     "vega": "github:vega/vega#v3.2.1",
     "vega-datasets": "^1.18.0",
-    "vega-lite": "github:vega/vega-lite#v2.3.0",
+    "vega-lite": "^v2.3.1",
```
domoritz: This doesn't work because of the examples.

ydlamba: Oh! What the possible solution then?

ydlamba:
I have tested it, config gets reset every time as view initializes i.e when a new visualization is loaded. If you think this won't work then applying it spec is the other possible solution.

package.json
domoritz: Use GitHub but use the latest version.

ydlamba: I've updated it to `github:vega/vega-lite#v2.3.1` & it's working fine.

domoritz:
I think it's a bit magical right now so I think we should apply the config until we have a way to display the config in the editor.

ydlamba:
@domoritz I've done required changes. You can check it now.

PR #68, opened 2018-05-15
Upgrade to webpack 4

ydlamba:
Currently, not correctly working for production. (Work in Progress)


package.json
```diff
@@ -80,8 +50,10 @@
     "vega-tooltip": "^0.9.16"
   },
   "scripts": {
-    "start": "node scripts/start.js",
-    "build": "node scripts/build.js",
+    "start": "webpack-dev-server --mode development --hot --inline --progress --open --color",
+    "build": "webpack --mode production --progress",
+    "clean": "rm -rf build",
+    "postbuild": "cp -rn public/* build",
```
domoritz: Why do we need a build folder and don't build directly into the public folder?

public/index.html
```diff
@@ -3,10 +3,7 @@
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
-    <meta name="theme-color" content="#325">
-
-    <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico">
```
domoritz: No more favicon?

domoritz:
- [ ] Delete deploy.sh

package.json
domoritz: Ahh, I see. Maybe better name it `deploy`.

domoritz:
The build works for me. Why do you say it doesn't work in production?

domoritz:
- [ ] Reduce the number of chunks. 47 seems excessive unless I'm missing why they are necessary.

package.json
ydlamba: `/dist: "distribution", the compiled code/library`
I think `dist` is like more standard.

public/index.html
ydlamba: Actually, `html-webpack-plugin` was loading it from `public` but now I think it's better to add the favicon in index.html itself. Adding it back.

domoritz:
I think we want a few chunks so that downloads can run in parallel but not 47. 

What's the difference between cp and rsync?

ydlamba:
There is not much difference between `cp` and `rsync` in copying local files. Just I have seen rsync as a preferred tool in many places. 
> Rsync is better since it will only copy only the updated parts of the updated file, instead of the whole file. It also uses compression and encryption if you want.

ydlamba:
I have limited the maximum number of chunks to 5 (I also tried with 10 but it was creating small chunks thereafter, not much helpful). Try it if that works.

PR #80, opened 2018-05-18
Update `.travis.yml` for automated deployment

### LINKED ISSUE #78: Automatic deployment

ydlamba:
**Update `.travis.yml` for automated deployment**

@domoritz You need to put your Github Token here. Then we need to test this.

Fixes #78.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,11 +14,12 @@ deploy:
   local-dir: dist
   name: Dominik Moritz
   email: domoritz@gmail.com
-  project-name: vega-editor
+  project-name: editor
   addons:
     apt:
       packages:
         - parallel
   on:
     branch: master
+    tags: true
   
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,18 +3,19 @@ node_js:
   - "8"
 script:
   - yarn run lint
-  - yarn run generate-example-images
   - yarn run build
 cache: yarn
 deploy:
   provider: pages
   skip-cleanup: true
-  github-token: $GITHUB_TOKEN  # Set in the settings page of your repository, as a secure variable (Over to you, dom)
+  github-token: $GITHUB_TOKEN
   keep-history: true
   local-dir: dist
   name: Dominik Moritz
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,8 +1,24 @@
...
 cache: yarn
+deploy:
+  provider: pages
+  skip-cleanup: true
+  github-token: $GITHUB_TOKEN  # Set in the settings page of your repository, as a secure variable (Over to you, dom)
+  keep-history: true
+  local-dir: dist
+  name: Dominik Moritz
+  email: domoritz@gmail.com
+  project-name: vega-editor
```
domoritz: Isn't the name vega/editor or just editor?

.travis.yml
```diff
@@ -1,8 +1,24 @@
...
 node_js:
   - "8"
-
 script:
   - yarn run lint
-
+  - yarn run generate-example-images
+  - yarn run build
 cache: yarn
+deploy:
```
domoritz: Do we need to say that we only want to do this on master and only for tagged commits?

.travis.yml
ydlamba: Let's make it by the repo name i.e. `editor`. (Although not necessary)

.travis.yml
ydlamba: Tagged commits makes more sense. Updating it.

.travis.yml
```diff
@@ -1,8 +1,24 @@
...
+  local-dir: dist
+  name: Dominik Moritz
+  email: domoritz@gmail.com
+  project-name: vega-editor
+  addons:
+    apt:
+      packages:
+        - parallel
+  on:
+    branch: master
```
domoritz: add `tags: true`

.travis.yml
domoritz: see below

.travis.yml
domoritz: Can you just delete it? It's not really necessary.

PR #83, opened 2018-05-19
Export as PDF feature

### LINKED ISSUE #82: Add a feature for exporting as PDF

ydlamba:
Fixes #82.

### DIFF
--- a/src/components/renderer/renderer.tsx
+++ b/src/components/renderer/renderer.tsx
@@ -1,5 +1,6 @@
 import './index.css';
 
+import jsPDF from 'jspdf';
 import * as React from 'react';
 import * as vega from 'vega';
 import vegaTooltip from 'vega-tooltip';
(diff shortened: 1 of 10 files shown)


src/components/renderer/renderer.tsx
```diff
@@ -58,12 +60,24 @@ export default class Editor extends React.Component<Props> {
...
       }).catch(err => {
         throw new Error('Error in exporting: '+ err);
       });
     }
+
+    if (props.exportPDF) {
+      const pdf = new jsPDF('p', 'mm', 'a4');
+      const url = Editor.view.toImageURL('png');
+      url.then(data => {
+        pdf.addImage(data, 'PNG', 0, 0);
```
domoritz: png is not vector so the pdf will not have high resolution

domoritz:
We should only support this in the SVG mode.

domoritz:
Closing as client side PDF generation doesn't work well.

PR #84, opened 2018-05-20
Update export visualization feature (Experimenting)

### LINKED ISSUE #56: SVG export doesn't work in Firefox
See https://github.com/vega/vega-lite/issues/3603

ydlamba:
This might fix #56.


domoritz:
You can't save the page anymore now. This means I cannot save the SVG.

screen shot 2018-05-19 at 17 48 25

ydlamba:
We can't even save the page on the deployed version as well. Try `Print` and then `Save as PDF`. That's the only option which I can think as of now.

domoritz:
Hmm, I was sure that we could do this in the past.

domoritz:
Nice, this works!

PR #92, opened 2018-05-21
Fix formatting spec bug

### LINKED ISSUE #88: Formatting bug
Hitting the format button multiple times formats again.

ydlamba:
Fixes #88.


src/components/input-panel/spec-editor/renderer.tsx
```diff
@@ -127,7 +127,6 @@ class Editor extends React.Component<Props, State> {
       provideDocumentFormattingEdits(model, options, token) {
         return [
           {
-            range: model.getFullModelRange(),
```
ydlamba: This is what a bug looks like. You never know where it is. :bug:

PR #97, opened 2018-05-28
Add fullscreen feature

### LINKED ISSUE #90: Full screen

ydlamba:
Fixes #90.
Right now, buttons have styling like that of toolbar buttons.


domoritz:
Looks great. Make the close button much larger, though. I also wonder whether we need to say that the close button goes back to the editor ("Edit in the Vega Editor"). Also, the URL should change when you're in fullscreen view.

ydlamba:
**Updates:**

- Add routes for fullscreen mode
- Add icons instead of texty buttons.
- Add tooltip over icons.

To check route in action. Open an example and then go to fullscreen mode. There is no change in URL for custom specs because they cannot be shared yet.

domoritz:
* Size of button should not be relative to screen size
* Large vizes need to be scrollable
* Zooming in the browser should work

ydlamba:
Bugs Fixed.

src/components/renderer/index.css
```diff
@@ -25,3 +25,41 @@
...
+  right: 0.8vw;
+  cursor: pointer;
+}
+.fullscreen-close {
+  position: absolute;
+  background: #eee;
+  border: 2px solid #000;
+  width: 200px;
+  height: 40px;
+  top: 2vw;
```
domoritz: Don't size things relative to the screen

src/components/renderer/renderer.tsx
```diff
@@ -16,12 +19,26 @@ interface Props {
   mode?: Mode;
   export?: boolean;
   baseURL?: string;
+  history?: any;
 }
 
-export default class Editor extends React.Component<Props> {
+interface State {
+  imageURL: string;
```
domoritz: don't use the image, use the vega view

src/components/renderer/renderer.tsx
```diff
@@ -89,14 +116,63 @@ export default class Editor extends React.Component<Props> {
...
+      <div>
+        <div className="chart">
+          <div ref="chart" />
+        </div>
+        <PortalWithState
+          closeOnEsc
+          onOpen={() => {
+            this.setState({ fullscreen: true });
+            const params = pathname.split('/');
+            if (pathname !== '/' && pathname !== '/edited' && params[params.length - 1] !== 'view') {
```
domoritz: is there a cleaner way to write this?

src/components/renderer/renderer.tsx
```diff
@@ -89,14 +116,63 @@ export default class Editor extends React.Component<Props> {
...
+        >
+          {({ openPortal, closePortal, portal }) => (
+            <React.Fragment>
+              <img data-tip="Fullscreen" className="fullscreen-open" onClick={openPortal} src="images/fullscreen.svg" />
+              {portal(
+                <div className="fullscreen-chart">
+                  <button className="fullscreen-close" onClick={closePortal}>
+                    <span>{'Edit it the Editor'}</span>
+                  </button>
+                  <img src={this.state.imageURL} />
```
domoritz: Use the vega view directly. Otherwise tooltips and stuff won't work

ydlamba:
**Updates:**

- The view is used instead of image in fullscreen
- Button/Icon sizes are in pixels instead of vw/vh.

P.S: Interactive states cannot be preserved in fullscreen mode because view instance can only use one container at a time. So changing the container is re-initializing the view.

PR #112, opened 2018-06-15
Adding Feature: Data Viewer

### LINKED ISSUE #85: Feature to view dataset used in a spec

ydlamba:
**First Draft**
screenshot_20180615_201616
**Status:** Work in Progress
Fixes #85.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "react-feather": "^1.1.1",
     "react-ga": "2.5.3",
     "react-monaco-editor": "^0.17.2",
+    "react-paginate": "^5.2.3",
     "react-portal": "^4.1.5",
     "react-redux": "^5.0.7",
     "react-router-dom": "^4.3.1",
--- a/src/actions/editor.ts
+++ b/src/actions/editor.ts
@@ -5,6 +5,7 @@ export const FORMAT_SPEC: 'FORMAT_SPEC' = 'FORMAT_SPEC';
 export const LOG_ERROR: 'LOG_ERROR' = 'LOG_ERROR';
 export const PARSE_SPEC: 'PARSE_SPEC' = 'PARSE_SPEC';
 export const SET_BASEURL: 'SET_BASEURL' = 'SET_BASEURL';
+export const SET_DATASETS: 'SET_DATASETS' = 'SET_DATASETS';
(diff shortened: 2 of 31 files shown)


domoritz:
Somehow I cannot brush in http://localhost:8080/#/examples/vega-lite/selection_layer_bar_month.

Broken hover

screen shot 2018-06-26 at 17 25 40

- [ ] make tabs look like in VSCode or remove the line between tab and content.
- [ ] dropdown too large
- [ ] Too much white space
- [ ] It's not just error logs. Rename!

screen shot 2018-06-26 at 17 26 15

Resize should be above the tabs

screen shot 2018-06-26 at 17 29 37

Other

- [ ] Don't use `(...)`
- [ ] How do I access data that's too large?
- [ ] Only show the first 100 rows. Then provide a way to jump to next batch.

ydlamba:
@domoritz I have updated a bunch of things. To proceed further, PR needs your feedback.

ydlamba:
The only thing left is `How do I access data that's too large?`, which will be done in a followup PR.

ydlamba:
PR has been updated.

domoritz:
When the table is narrow, the pager is oddly placed. Do you want to center it?

screen shot 2018-08-01 at 08 57 06

domoritz:
Before we can merge this, we need to deal with `...`. At least we should show a tooltip that says that the field is too large to be displayed. Please use the view API (see JS console).

domoritz:
Sometimes headers are not fully visible

screen shot 2018-08-01 at 09 09 00

domoritz:
- [ ] Data table does not update when the data changes. Doe example, when I brush in a chart, the brush store should update.

domoritz:
- [ ] The state of the data viewer is not saved. When I refresh, the arrow is pointing the wrong way.

domoritz:
- [ ] I think the header of the table should not scroll.

screen shot 2018-08-01 at 09 12 30

domoritz:
- [ ] Large table names overflow the dropdown.

screen shot 2018-08-01 at 09 13 12

domoritz:
Two things and then we can merge

- [ ] horizontal scrolling when the table is too wide
- [ ] use the full height for the table instead of fixing the height

PR #113, opened 2018-06-20
Upgrade to Vega 4 rc-3

ydlamba:


domoritz:
Why have all the images changed?

ydlamba:
Running `yarn generate-example-images` changed them. They shouldn't change, right? It might be a problem with my setup.

ydlamba:
I didn't generate the images as all of them are changing for me. You need to do that.

domoritz:
Okay. They will be built on travis so let's just leave it for now.

PR #120, opened 2018-06-23
Mode awareness

### LINKED ISSUE #117: Mode awareness
When we are in Vega-Lite mode, we should open the gist and the example gallery with Vega-Lite preselected.

ydlamba:
Fixes #117.

### DIFF
--- a/src/components/header/renderer.tsx
+++ b/src/components/header/renderer.tsx
@@ -89,7 +89,7 @@ class Header extends React.Component<Props & { history: any }, State> {
     this.props.history.push('/custom/vega-lite');
   }
   public onClear() {
-    if (this.props.mode === 0) {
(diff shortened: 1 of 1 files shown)


src/components/header/renderer.tsx
```diff
@@ -49,7 +49,7 @@ class Header extends React.Component<Props & { history: any }, State> {
       gist: {
         filename: '',
         revision: '',
-        type: 'vega',
+        type: props.mode === Mode.Vega ? 'vega' : 'vega-lite',
```
domoritz: why not just `props.mode`?

ydlamba: `props.mode` is of type number.

ydlamba: Done!

PR #126, opened 2018-07-03
Adding Feature: Snapshot

ydlamba:
**To Do -**
- [x] Add button in export modal
- [x] Add copy to clipboard option
- [x] Show generated URL in modal
...


src/components/app-shell.tsx
```diff
@@ -21,6 +21,7 @@ export default class AppShell extends React.Component<Props> {
           <Route path="/gist/:mode/:username/:id/:revision/:filename" component={App} />
           <Route path="/examples/:mode/:example_name" component={App} />
           <Route path="/custom/:mode" component={App} />
+          <Route path="/shared/:spec" component={App} />
```
domoritz: How do we know whether it's Vega or Vega-Lite?

src/components/header/renderer.tsx
```diff
@@ -248,6 +249,12 @@ class Header extends React.Component<Props & { history: any }, State> {
         {'Export'}
       </div>
     );
+    const saveButton = (
+      <div className="header-button" onClick={() => this.props.saveSpec(true)}>
+        <Save className="header-icon" />
+        {'Snapshot'}
```
domoritz: Wasn't the plan to put this in the sharing modal?

src/components/app-shell.tsx
domoritz: rename to just `share`

domoritz:
- [ ] Add compression

ydlamba:
Pull request has been rebased and Now, I'm working on the snapshot feature.

domoritz:
- When the sharing modal is open, always generate the URL. Don't wait. 
- Instead of a text input, let's use a URL so the user can see the changes when they check fullscreen. Preview is then just a matter of clicking on the button
- No need to the big button. 
- make the copy to clipboard more visible. Add some text for it! Never just rely on icons.

domoritz:
- add a byte counter (and show a warning when we are over 2000)
- make the link look like a link. use underline and remvoe the grey box
- but the copy button in a prominent location
- write "copy to clipboard". copy by itself is not enough

src/components/header/renderer.tsx
```diff
@@ -479,6 +525,64 @@ class Header extends React.Component<Props, State> {
       </div>
     );
 
+    const copiedClass = 'copied' + (this.state.copied ? ' visible' : '');
```
domoritz: Use string template

src/components/header/renderer.tsx
```diff
@@ -479,6 +525,64 @@ class Header extends React.Component<Props, State> {
...
+            </div>
+          </Clipboard>
+          <div className={copiedClass}>Copied!</div>
+        </div>
+        <div className="byte-counter">
+          Bytes: {this.state.generatedURL.length}{' '}
+          <span className="warning">
+            {this.state.generatedURL.length > 2083 ? (
+              <span>
+                Warning:{' '}
```
domoritz: add space between : and content

src/components/input-panel/spec-editor/renderer.tsx
```diff
@@ -90,7 +91,8 @@ function debounce(func, wait, immediate?) {
 interface Props {
   autoParse?: boolean;
   format?: boolean;
-  history;
+  history: any;
+  match: any;
```
domoritz: what's this for?

src/components/header/index.css
```diff
@@ -433,3 +527,18 @@
     margin-right: -105px;
   }
 }
+
+/* Scrollbar Styles */
```
domoritz: These are global styles, right? Move them somewhere else then.

src/components/header/index.css
```diff
@@ -78,6 +78,7 @@
   position: relative;
   background: white;
   width: 80%;
+  font-family: "Roboto", "Helvetica", "Arial", sans-serif;
```
domoritz: Why is this necessary?

src/components/app.css
```diff
@@ -77,6 +77,13 @@ html {
   overflow: visible !important;
 }
 
+button,
+input,
+textarea,
+select {
+  font-family: "Roboto", "Helvetica", "Arial", sans-serif;
```
domoritz: Again? I thought we set this on the body.

src/components/header/renderer.tsx
ydlamba: There is a space in the final rendering.

ydlamba:
PR has been updated with final changes.

src/components/header/index.css
domoritz: They are still here

src/components/header/index.css
```diff
@@ -413,6 +403,110 @@
...
+  vertical-align: middle;
+  line-height: 25px;
+  white-space: nowrap;
+}
+
+.exported-url a::selection {
+  background: rgba(0, 0, 0, 0.15);
+}
+
+.share-buttons,
```
domoritz: We are still not sharing the style between the data viewer and the modals, right?

PR #127, opened 2018-07-03
Bug Fix: Cross-origin error

### LINKED ISSUE #123: A cross-origin error was thrown in editor
There is an error "Error: A cross-origin error was thrown. React doesn't have access to the actual error object in development. See https://fb.me/react-crossorigin-error for more information."

When giving the editor this spec:
```
{
  "$schema": "https://vega.github.io/schema/vega/v4.json",
  "width": 800,
  "height": 500,
  "data": [{"name": "drive", "url": "data/driving.json"}],
  "scales": [
    {
      "name": "x",
      "type": "linear",
      "domain": {"data": "drive", "field": "miles"},
      "range": "width"
    },
    {
      "name": "y",
      "type": "linear",
      "domain": {"data": "drive", "field": "gas"},
      "range": "height"
    }
  ],
  "marks": [
    {
      "type": "text",
      "from": {"data": "drive"},
      "encode": {
        "enter": {
          "x": {"scale": "x", "field": "miles"},
          "y": {"scale": "y", "field": "gas"},
          "text": {"field": "year"}
        }
      },
      "transform": [{"type": "filter", "expr": "datum.mile > 10"}]
    }
  ]
}
```

ydlamba:
Fixes #123.
Replacing `cheap-module-eval-source-map` with `cheap-module-source-map` will fix the issue but increase the rebuilding time during development.

Reference: https://webpack.js.org/configuration/devtool/#devtool and https://reactjs.org/docs/cross-origin-errors.html

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -17,7 +17,7 @@ module.exports = (env, argv) => {
 
     devtool:
       argv.mode === "development"
-        ? "cheap-module-eval-source-map"
+        ? "cheap-module-source-map"
         : "source-map",
 
     resolve: {


ydlamba:
https://reactjs.org/docs/cross-origin-errors.html#webpack says - 

> Some JavaScript bundlers may wrap the application code with eval statements in development. (For example Webpack will do this if devtool is set to any value containing the word “eval”.) This may cause errors to be treated as cross-origin.

>If you use Webpack, we recommend using the cheap-module-source-map setting in development to avoid this problem.

kanitw:
Nice! I was running into this a few times recently. :)

PR #133, opened 2018-07-12
Add autoprefixer and speed up webpack builds

ydlamba:


domoritz:
Where are you disabling garbage collection?

What problem does adding auto prefixer solve?

ydlamba:
>Where are you disabling garbage collection?

screenshot_20180712_205144
In case you want to go through the blog: https://medium.com/@kenneth_chau/speeding-up-webpack-typescript-incremental-builds-by-7x-3912ba4c1d15

>What problem does adding auto prefixer solve?

`autoprefixer` solves the minor styles issues across different browser and their versions. For example, `user-select: none` works great in Chrome but not in Firefox or Safari. There are few very small style issues which hardly any user notices but ideally they shouldn't be there.

Issue #33, opened 2018-02-21
Use JSON request service


domoritz:
right now, we hard code which schemas are used for completion. We completely ignore the schema from the spec. This is not good. 

To understand this issue, use a specific version of the schema and note how completion doesn't work anymore.

nesterone:
Hi, @domoritz  
I see it at https://github.com/vega/vega-editor/blob/bfb3c59d03cdbbacdf6a8a4639767121e7aaa644/src/utils/validate.js#L10

If you don't mind I can take this one, play around and send PR for review

nesterone:
After a playing with validator I can propose next solution:
- put schemas in redux store (for example with key as url) 
- provide separate redux action `UPDATE_VEGA_SCHEMA` and `UPDATE_VEGA_LITE_SCHEMA`
- keep `validationSchema` in actions which trigger validation 
- map current schema to the component through connect

domoritz:
Thank you for looking into this. I'm not sure I understand why the schema needs to be stored in the redux store. This issue is about the json request service that is available in vscode but not by default in Monaco.

nesterone:
Looks like I've mixed up validation with autocomplete. Let me take a look on how monaco editor manages autocomplete ... 

P.S.: 

Is it ok that editor won't take to account vega spec`s schema during validation (as far as I see validation is hardcoded )

domoritz:
Yeah, the autocomplete and validation in the editor component are separate from the validation outside the editor component. Ideally, they would be together but I haven't gotten round to fixing it. Thanks for looking into this!

nesterone:
Hi, @domoritz 
I have found a way to make Monaco editor to load schemas on-demand 

https://monosnap.com/file/hlvnvQQCFUedZFb3jDm93aaPe1b4nJ

It's based on the same functionality as VSCode

However, it  requires to modify Microsoft/monaco-json. My current changes here https://github.com/nesterone/monaco-json/commits/custom-request-service

If you don't mind I'm going to make a pull request to Microsoft/monaco-editor with my changes and references to that issue in vega/editor

domoritz:
Oh sweet. Thanks for looking into this. Since we will still only load the latest vega or vega-Lite version, we will have to show a warning if the versions are not compatible but that should be easy. 

[...]

PR #135, opened 2018-07-17
[WIP] Use JSON request service #33

nesterone:
- brute force schema preloading from spec


domoritz:
Thank you for the PR. The issue with this change is that now the schema validation will be inconsistent between what monaco does and what the external validation does. We should make sure that it's the same.

src/components/input-panel/spec-editor/renderer.tsx
```diff
@@ -183,14 +132,14 @@ class Editor extends React.Component<Props, {}> {
   public componentWillUnmount() {
     document.removeEventListener('keydown', this.handleKeydown);
   }
-  public updateSpec(spec) {
+  public async updateSpec(spec) {
     let parsedMode = this.props.mode;
 
     try {
-      const schema = JSON.parse(spec).$schema;
+      const schemaUrl = JSON.parse(spec).$schema;
```
domoritz: I wonder whether we can put the logic below somewhere else where we have already parsed the spec into an object.

nesterone:
Looks like I've missed slack channel, joining ...

domoritz:
Having said that, github issues are great because we keep a history.

nesterone:
Yep, let's continue the issue related discussion here
So you are proposing to investigate API of json service in monaco editor for on-demand loading of schemas, I'm not so optimistic but it worse to try

domoritz:
Yes, that would be nice. Vscode uses it so it ought to work but maybe it replies on some node stuff.

domoritz:
Sorry for being so pedantic. Are you still working on this?

nesterone:
Yep, right now I'm looking into the code of https://github.com/Microsoft/vscode-json-languageservice and trying to understand on how it was extracted from VSCode.
There is a big chance that functionality with on-demand loading for schemas was left in VS code base

PR #146, opened 2018-08-06
Follow-up Fixes: Data-viewer

ydlamba:


src/components/data-viewer/renderer.tsx
```diff
@@ -88,7 +88,11 @@ export default class DataViewer extends React.Component<Props, State> {
 
     const visibleData = data.slice(start, end);
 
-    const table = data.length ? <Table header={Object.keys(data[0])} data={visibleData} /> : <div>Empty table</div>;
+    const table = data.length ? (
+      <Table header={Object.keys(data[0])} data={visibleData} />
+    ) : (
+      <span className="error">No table found. Refresh button might help you.</span>
```
domoritz: replace with "Try to refresh."

src/components/data-viewer/renderer.tsx
```diff
@@ -109,7 +113,7 @@ export default class DataViewer extends React.Component<Props, State> {
           </button>
           <div className="pagination-wrapper">{pagination}</div>
         </div>
-        <ErrorBoundary>{table}</ErrorBoundary>
+        {this.props.error ? <span className="error">Something is wrong. Check logs.</span> : table}
```
domoritz: This is not a clean solution. The error boundary was good. Please revert this.

domoritz: Let's keep the error boundary.

domoritz: Instead of this error message, just open the error log automatically. The data viewer will never be useful when there are errors. For warnings, I propose https://github.com/vega/editor/issues/147

PR #155, opened 2018-08-12
Bug Fix: Wrong sharing URL

ydlamba:


ydlamba:
@domoritz The generated URL on the deployed editor is missing something which opens the wrong page. This is a fix for that.

PR #161, opened 2018-09-09
Bug Fix: Save parsing option

### LINKED ISSUE #160: Manual parsing selection is not saved
When I reload the editor, the setting gets reset to auto parsing.

ydlamba:
Fixes #160.


domoritz:
I still can't use the button in auto parse, can I?

ydlamba:
I'm not sure why we want to click on the run button while parsing is auto?

Issue #148, opened 2018-08-06
Automatically open logs when there is an error


domoritz:
We never want to see the data viewer instead.

ydlamba:
We can disable the data-viewer tab if there is an error.

domoritz:
Done in #163

PR #163, opened 2018-09-18
Improve debug pane

ydlamba:
Automatically switch tab to logs and disable data-viewer if there is an error.
#148.


ydlamba:
Should we also automatically open the debug pane in case of errors?

domoritz:
Yes, open it for errors as well.

domoritz:
Thank you. Looks great.

PR #180, opened 2018-11-15
Use JSON request service

### LINKED ISSUE #33: Use JSON request service

nesterone:
- upgrade monaco-editor
- turn on schema request prop
- close #33


domoritz:
Replaced by https://github.com/vega/editor/pull/181

PR #202, opened 2019-01-23
feature / vega lite + gist: relative data source URLs  become absolute ones for existing files

martinvirtel:
This is about using the editor to display a vega-lite spec from a gist.

A spec that contains a Vega Lite spec with a "url"-type data source currently can be viewed in the editor only if the URL is absolute (or can be successfully resolved if the data file exists under the same hostname as the editor). This was preventing me from using relative urls in specs that are stored in gists. 

I changed the handling of gists so that if there is a relative URL as data source and the referenced file exists in the gist the data source will be updated to a absolute url.

Here's an example of such a gist:

https://gist.github.com/martinvirtel/556bc74cfdf07e194fe609da28c2aebd/#file-spec-relative-vl-json

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@ services:
     init: true
     build: .
     ports:
-      - "18060:8080"
+      - "8080:8080"
     volumes:
       - "./public:/usr/src/app/public"
       - "./src:/usr/src/app/src"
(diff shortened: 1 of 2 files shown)


martinvirtel:
Forgot to lint, sorry, closing.

Issue #205, opened 2019-02-14
save or export  as .vl.json and .vg.json (feature request)


iliatimofeev:
Add an option in export to save as vega-lite and vega specifications.

domoritz:
PRs very welcome.

siddhant1:
Closing since fixed

PR #210, opened 2019-02-28
add feature: export specifications in json format

Esorakouki:
add feature #205
add-json-export


domoritz:
Thank you. We'll review this and get back to you soon.

siddhant1:
@domoritz Have you testesed this ? This looks promising .
Should I test this once on my machine?

domoritz:
This looks great! @siddhant1 that would be fantastic. 

We need to change something about the layout of these buttons soon. They should probably wrap instead of being one long row. But that's a separate PR.

siddhant1:
Sure , that'll be a great improvement !

domoritz:
One thing at a time.

domoritz:
@siddhant1 Can you finish this so we can merge it?

siddhant1:
Is there any change remaining in this one? 
I will test i today on my machine.
Also , should I make all the buttons in different rows?

domoritz:
Rebase and test. I also don't think we need to change the width in css as done in this PR. Let's get this merged before making the export actions multiple rows. But it should be done, yes.

siddhant1:
Tested ! Worked Great for me

src/components/header/renderer.tsx
```diff
@@ -242,6 +249,28 @@ class Header extends React.Component<Props, State> {
     dlButton.classList.remove('disabled');
   }
```
domoritz: No need to be async

src/components/header/index.css
```diff
@@ -399,7 +399,7 @@
 
 .export-button {
   flex: 1;
-  min-width: 280px;
+  min-width: 270px;
```
domoritz: Please undo

src/components/header/index.css
```diff
@@ -411,7 +411,7 @@
   cursor: pointer;
   user-select: none;
   padding: 10px 20px;
-  margin-right: 40px;
+  margin-right: 30px;
 }
```
domoritz: Please undo

domoritz:
I left a review. Please update and then we can merge

domoritz:
We still use async.

siddhant1:
Yeah , I had a doubt about it , I mentioned it above

domoritz:
Where?

siddhant1:
https://github.com/vega/editor/issues/237#issuecomment-470121959

siddhant1:
https://github.com/vega/editor/pull/210#discussion_r263005659

siddhant1:
So my doubt was , Doesn't this.props.view.ToImageURL returning promise which he is awaiting via await , I think thats why he made the function async .

src/components/header/renderer.tsx
domoritz: Yes, but we don't need it for json.

siddhant1: Yes , I got it and pushed the changes too , this was a previous fork so there are some linting issues , I will fix them in my next pr.

src/components/header/renderer.tsx
```diff
@@ -242,6 +249,28 @@ class Header extends React.Component<Props, State> {
...
+    if (
+      event.target &&
+      (event.target.matches('input') ||
+        event.target.matches('label') ||
+        event.target.matches('div.type-input-container'))
+    ) {
+      return;
+    }
+    const content = this.state.downloadVegaJSON ? this.props.vegaSpec : this.props.vegaLiteSpec;
+    const filename = this.state.downloadVegaJSON ? 'visualization.vg.json' : 'visualization.vl.json';
```
domoritz: In a follow up, please use template strings.

domoritz:
Merge when test pass.

src/components/header/renderer.tsx
siddhant1: Do they serve any benefit here?

domoritz: Yes, more readable code.

Issue #220, opened 2019-03-02
Accidentally closing editor while editing


ayushnawal:
While playing or editing with the json data , if the editor gets accidentally closed by the user , a warning must be generated before closing the window to save the unsaved changes , as user's last edit will not be preserved automatically .

I have solved issue and now our editor will ask you to save changes you have edited before closing the window .
The "Leave Site ? " dialog box will only appear if you have edited some data in the input panel .

I am attaching a screenshot how it will appear if you have edited something : 

![screenshot from 2019-03-03 04-42-36](https://user-images.githubusercontent.com/36100944/53688790-47fdf280-3d6f-11e9-8788-c650bbd331f5.png)

@domoritz if you don't mind , can i send my PR , so that you can review it :)

domoritz:
> as user's last edit will not be preserved automatically

We actually do save the last state. I think we want this dialog only if there are unsaved changes for example when you have manual parsing on.

ayushnawal:
@domoritz i was actually trying to convey that only if the user have any unsaved changes , than a pop-up should appear before the editor's window gets closed accidentally . I think every editor generate a warning before leaving the site with any unsaved changes .

domoritz:
Go ahead and send a PR to fix this issue. I agree that we should behave like any other editor.

domoritz:
We are actually saving the editor content and preserve it across refreshes.

PR #221, opened 2019-03-03
Adding leave-site? alert box

ayushnawal:
The site did not show any alert box when a user accidentlly closes the editor with
unsaved changes.

Now if he edits manually in the input panel and the editor get closes by mistake
the alert box will warn him that any unsaved changes will be lost.

Fixes https://github.com/vega/editor/issues/220


src/components/input-panel/spec-editor/renderer.tsx
```diff
@@ -102,6 +102,10 @@ class Editor extends React.Component<Props, {}> {
 
     if (this.props.history.location.pathname.indexOf('/edited') === -1) {
       this.props.history.push('/edited');
+      window.onbeforeunload =e => {
```
domoritz: Did you try this? I think this always catches, not just when we have unsaved changes. Please test carefully.

ayushnawal: @domoritz  Sir , you are right ! . It will alert the user whether the user have unsaved changes or not . I thought that we have alerted the user that  " Any unsaved changes will be lost? " and user will handle this query manually as he will also get assured also once if he/she have saved the changes . If you want dialog box only to appear when the changes are unsaved , I will try to FIX it asap . But Please don't mind i was thinking we should ask atleast once before leaving the editor as every editor site does .

domoritz: I see. I think it's better if we only show a warning if the user actually loses information if they navigate away. Thanks for looking into this!

ayushnawal: @domoritz ok i figured out that how we are storing last state and if we close the editor , on again opening the editor it renders last state in input panel . But it has led us to a new bug which is with the "Browser Back Button" . Let's say a user edited twice and than he jumps back to his previous edit using back button , there is no way he can get his edited code back by using forward button . As you know we are only storing the last state and it gets updated on every edit , we are not preserving history of multiple edits . i hope you are getting what i am trying to convey . Thanks for consideration !

domoritz: Yes, there are numerous usability issues with back and forward buttons but fixing them requires some major changes that we have not done yet. You're more than welcome to tackle this but it's not an easy task.

ayushnawal: Yes , its seems to be a tricky task , i will keep trying if i figured out any solution , i will try to keep you updated on this bug !

domoritz:
Please open a new PR.

Issue #225, opened 2019-03-03
Invalid URLs


ayushnawal:
On Navigating any invalid url with editor , no 404 component is rendering .

Till now when we type any invalid url for example : https://vega.github.io/editor/#/abc

it's displaying a blank page (not rendering anything) which should not be the case as it is quite confusing !

@domoritz  if you don't mind i am resolving this issue and will send a PR asap .

Thanks!

domoritz:
I don't think this is a huge deal.

PR #226, opened 2019-03-03
Adding 404 component

ayushnawal:
Adding page not found component which will render 404 error page on navigating invalid
urls . Till now a blank page was rendering .

Fixes https://github.com/vega/editor/issues/225

@domoritz sir , please review it .

Thanks !


ayushnawal:
Okay i am changing font-family , Do i need to change anything more ?

domoritz:
Actually, see the pr to add some new routes. 

Why don't we just you're people to the main app and show a toast message?

ayushnawal:
I think i am not getting you correctly . Are you saying on navigating through invalid urls , we should bring user back to the editor's landing page ?

domoritz:
I think that's the best user experience, no?

ayushnawal:
If you don't mind i am think this will be a more better way that : on navigating on an invalid URL , we will render the 404 component but we will also add a line on the page like "Go back to the editor : (link) " or something like that .
What do you think ?

domoritz:
To be honest, nobody has complained about this issue before so I would like to do the minimal thing. How about we only redirect to the main page and that is it?

ayushnawal:
@domoritz we are rendering main page on invalid urls now as you suggested .
Please review . 
I think we can close this issue now .
Thanks :)

src/components/app-shell.tsx
```diff
@@ -17,11 +18,15 @@ export default class AppShell extends React.Component<Props> {
       <div>
         <Switch>
           <Route path="/" exact component={App} />
-          <Route path="/edited" component={App} />
-          <Route path="/gist/:mode/:username/:id/:revision/:filename" component={App} />
-          <Route path="/examples/:mode/:example_name" component={App} />
-          <Route path="/custom/:mode" component={App} />
-          <Route path="/url/:mode/:compressed" component={App} />
+          <Route exact path="/edited" component={App} />
```
domoritz: What does exact change here?

ayushnawal: The exact param here disables the partial matching for a route and makes sure that it only returns the route if the path is an EXACT match to the current url. 

Actually i added this when i was rendering a 404 component on routes which are not defined , so i added exact param.

domoritz: what is partial matching? Can you give me an example?

domoritz:
Please merge master. Thanks!

src/components/app-shell.tsx
ayushnawal: we are making use of Switch Component, which renders the first route that matches, Router doesn't look for a complete match .

```
<Route path="/home" component={home} />
<Route path="/home/room" component={room} />

```
when we go to http://app.com/home , no problem occurs and everything works fine .
But, if we went to http://app.com/home/room, it would again go through all of our defined routes and return the FIRST match it finds. React router does partial matching, so` /home `partially matches `/home/room`.

So, the exact param disables this partial matching

domoritz: I see. Thanks.

domoritz:
Please redo this PR.

Issue #228, opened 2019-03-03
Report an issue


ayushnawal:
We should add a small Report Bug button in editor itself , it will help user to directly communicate his problems with community and like every other open source project does, it will help in inviting contributors for enhancement of the project !

@domoritz i have resolved this issue . Please Review .

Thanks !

siddhant1:
Since this is fixed ,we can close this issue

Issue #217, opened 2019-03-02
Enhancement : Add contributing guidelines


siddhant1:
Since this project is coming is gsoc-2019, I think we should add contributing and raising issues guidelines.

domoritz:
Yeah, probably mostly linking to the Vega-Lite guidelines.

siddhant1:
I was thinking of adding pull request templates alongside too.
Like



#### Nature of issue?



- Found a bug
- Existing feature enhancement
- New feature request


#### Details about the bug: 

- Web browser and version: 
- Operating System: 
- Steps to reproduce this bug:




#### Feature enhancement details:



#### New feature details:


and pr templates


I have verified that this pull request:

* [ ] has no linting errors (`npm run lint`)
* [ ] is from a uniquely-named feature branch and has been rebased on top of the latest master. (If I was asked to make more changes, I have made sure to rebase onto master then too)
* [ ] is descriptively named and links to an issue number, i.e. `Fixes #123`

domoritz:
Thanks for pushing forward. I think a contributing guide is great! I don't think we need PR templates but issue templates would be great. I would also love to have a link in the editor to directly report a bug similar to what https://prettier.io/playground/ has in the bottom right. 

<img width="370" alt="screen shot 2019-03-02 at 10 11 38" src="https://user-images.githubusercontent.com/589034/53685742-9d84c180-3cd3-11e9-99e9-42aebc2e74ec.png">

siddhant1:
Yes , I think this is a great improvement I will surely work on this later today

PR #229, opened 2019-03-03
Adding Report Bug Button

ayushnawal:
Adding a report issue button in the toolbar to directly communicate with the user's
problem . it will also let us invite new contributers , which can help in further
enhancement.

Fixes https://github.com/vega/editor/issues/228

@domoritz Please review PR for changes (if any )

Thanks :)

### DIFF
--- a/src/components/toolbar/renderer.tsx
+++ b/src/components/toolbar/renderer.tsx
@@ -39,6 +39,14 @@ export default class Toolbar extends React.Component<Props> {
     return (
       <div className="toolbar">
         {this.showErrorAndWarnings()}
+        <div
+          className="report"
+          onClick={() => {
+            location.href = 'https://github.com/vega/editor/issues/new';
+          }}
+        >
+          REPORT ISSUE
+        </div>
         <div className="status">{`${NAMES[this.props.mode]} version ${getVersion(this.props.mode)}`}</div>
         <div
           className="renderer-toggle"
--- a/src/components/toolbar/renderer.tsx
+++ b/src/components/toolbar/renderer.tsx
@@ -45,7 +45,21 @@ export default class Toolbar extends React.Component<Props> {
             location.href = 'https://github.com/vega/editor/issues/new';
           }}
         >
-          REPORT ISSUE
+          Report Issue
+        </div>
+        <div
+          className="report"
+          onClick={() => {
+            const dummy = document.createElement('input');
+            const text = window.location.href;
+            document.body.appendChild(dummy);
+            dummy.value = text;
+            dummy.select();
+            document.execCommand('copy');
+            document.body.removeChild(dummy);
+          }}
+        >
+          Copy Link
         </div>
         <div className="status">{`${NAMES[this.props.mode]} version ${getVersion(this.props.mode)}`}</div>
         <div
--- a/src/components/toolbar/index.css
+++ b/src/components/toolbar/index.css
@@ -24,6 +24,12 @@
   cursor: default;
 }
 
+.report {
+  background: #000;
+  color: white;
+  cursor: default;
+}
+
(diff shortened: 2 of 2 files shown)


src/components/toolbar/renderer.tsx
```diff
@@ -39,6 +39,14 @@ export default class Toolbar extends React.Component<Props> {
     return (
       <div className="toolbar">
         {this.showErrorAndWarnings()}
+        <div
+          className="report"
+          onClick={() => {
+            location.href = 'https://github.com/vega/editor/issues/new';
+          }}
+        >
+          REPORT ISSUE
```
domoritz: let's not do all caps

src/components/toolbar/renderer.tsx
```diff
@@ -39,6 +39,14 @@ export default class Toolbar extends React.Component<Props> {
     return (
       <div className="toolbar">
         {this.showErrorAndWarnings()}
+        <div
+          className="report"
+          onClick={() => {
+            location.href = 'https://github.com/vega/editor/issues/new';
```
domoritz: We should include the spec itself here. Ideally, we have a link to the example.

Moreover, we should send an issue to the Vega and vega-Lite repos, not the editor, don't you think?

src/components/toolbar/renderer.tsx
ayushnawal: okay .. i will update it

domoritz:
I imagined something like https://github.com/vega/editor/issues/217#issuecomment-468945412

siddhant1:
I was working on it as of #217  , please comment on issues if you are working ! Thanks

src/components/toolbar/renderer.tsx
ayushnawal: > We should include the spec itself here. Ideally, we have a link to the example.

Sorry , i am not unable to get your point . Please clarify if you don't mind .

I think for now we should keep editor repo's link only , because they are interacting with the editor here and most issues will be related with interface / UI / UX of editor only according to me.

domoritz:
@siddhant1 Thanks for the note. There was already a discussion about this feature in https://github.com/vega/editor/issues/217#issuecomment-468945412 but the title of the issue was not very descriptive so I understand that @ayushnawal created a new issue and sent this PR.

ayushnawal:
Sorry @siddhant1  , I thought of this issue and than i checked all issues main title  , i did'nt looked every issue by opening and neither #217 , This issue was titled as Add contributing guidelines so i did'nt check that out  .

domoritz:
This happens. Thanks for understanding both of you!

ayushnawal:
@domoritz  i have updated the toolbar as you wanted .

screenshot from 2019-03-04 01-03-58

Adding more links or buttons in toolbar might make it look tedious .
I have also updated css for showing onclick affects on buttons 

Please Review . Thanks !

src/components/toolbar/renderer.tsx
domoritz: I doubt that. Most issues are related to the spec. Maybe instead of just a button, what we need are 

* A link to the Vega slack and Stack Overflow to ask questions
* A link to report an issue with the editor
* A button to report an issue with Vega or Vega-Lite, which includes a link to the spec with the sharing functionality from the editor. See how we create URLs there.

algomaster99: @domoritz One layout I can suggest is this. The color scheme will subject to change to though.
Screenshot from 2019-03-10 11-02-33
Please give your thoughts over it and suggest if changes are required.

domoritz:
Let's add this to the help modal I just merged.

domoritz:
We now have this in the modal.

Issue #232, opened 2019-03-04
Gist url validation


ArexJambusarwala:
Entering an invalid gist URL and pressing load doesn't do anything. It would be better to show an error message so that the user knows something has actually happened.
![issue14](https://user-images.githubusercontent.com/35253743/53711219-1bc59d00-3e67-11e9-8e51-6ad28b05132f.png)
A simple check on fetch can do the job.
Should I proceed with it?

domoritz:
How do you plan to show the error?

ArexJambusarwala:
There is already an error message but it is never displayed because the state is never changed. If the "invalidUrl" is changed this is displayed:
![issue15](https://user-images.githubusercontent.com/35253743/53723272-66f2a680-3e8d-11e9-91d7-5ff6d0e7c121.png)

PR #233, opened 2019-03-04
Gist url validated

ArexJambusarwala:
This pr will only check "Gist Url" field and not Revision and filename. The latter two work as they currently are. Also when the input url was faulty the url would still run. With this pr faulty urls wont run.


ArexJambusarwala:
Issue #232

src/components/header/renderer.tsx
```diff
@@ -153,10 +153,19 @@ class Header extends React.Component<Props, State> {
 
     const gistUrl = new URL(url, 'https://gist.github.com');
     const [username, gistId] = gistUrl.pathname.split('/').slice(1);
-
+    const gistCommits = await fetch(`https://api.github.com/gists/${gistId}/commits`).then(r => {
+      if (r.ok) {
```
domoritz: Please refactor. Set the state first outside the conditional.

src/components/header/renderer.tsx
```diff
@@ -170,20 +179,21 @@ class Header extends React.Component<Props, State> {
...
+        },
 
-      invalidUrl: false,
-    });
+        invalidUrl: false,
+      });
 
-    closePortal(); // Close the gist modal after it gets load
+      closePortal(); // Close the gist modal after it gets load
+    }
```
domoritz: Else? How does the user get feedback? Shouldn’t we show an error message?

domoritz: Nm, found it

ArexJambusarwala: There is an error message
corrected7
The message was already there before the pr but it wasn't being displayed.
Now it gets displayed after the load button is hit.

src/components/header/renderer.tsx
ArexJambusarwala: Will do.

domoritz: Lol, I mean, set the state to r.ok

src/components/header/renderer.tsx
```diff
@@ -153,10 +153,15 @@ class Header extends React.Component<Props, State> {
 
     const gistUrl = new URL(url, 'https://gist.github.com');
     const [username, gistId] = gistUrl.pathname.split('/').slice(1);
-
+    const gistCommits = await fetch(`https://api.github.com/gists/${gistId}/commits`).then(r => {
```
domoritz: One last thing. We can remove the `then` here and replace it with another await.

src/components/header/renderer.tsx
```diff
@@ -166,24 +171,28 @@ class Header extends React.Component<Props, State> {
       filename = Object.keys(gistData.files).find(f => gistData.files[f].language === 'JSON');
 
       if (filename === undefined) {
+        this.setState({
+          invalidUrl: true,
+        });
         throw Error();
```
domoritz: Do we still need this error? Would it make sense to show the user a bit more details about the error?

src/components/header/renderer.tsx
domoritz: Thanks

src/components/header/renderer.tsx
ArexJambusarwala: If you are talking about "Error()" then i believe it can removed!

domoritz: Well, we need it or otherwise we continue below, no? You would need to refactor the code with else cases. Yes, please do it.

src/components/header/renderer.tsx
ArexJambusarwala: It's done.

src/components/header/renderer.tsx
ArexJambusarwala: Consider this case:
corrected8
It gets loaded even though no file named dwdwd is present. This pr resolved issues with Gist-Url. If possible can I make a separate pr for this issue sir?

domoritz: Sure

PR #236, opened 2019-03-05
Validating revision and filename fields

### LINKED ISSUE #234: Show error message if filenames/revision do not match
If the filenames donot match show an error like:
issue16

If both Url and filename are incorrect show both errors as in:
corrected9

ArexJambusarwala:
This pr does the following:
=> Shows different error messages for Url, revision and filename.
=> Shows error messages below their respective fields.
=> Shows corresponding error if revision is invalid.
=> Shows corresponding error if filename is invalid.
Fixes #234 
Showing one possible case of correct Url but incorrect revision & filename:
corrected10

@domoritz Please review and suggest changes in error messages/refactoring.


ArexJambusarwala:
@domoritz Its done.

src/components/header/renderer.tsx
```diff
@@ -157,14 +161,31 @@ class Header extends React.Component<Props, State> {
     this.setState({
       invalidUrl: !gistCommits.ok,
     });
+    if (filename.length === 0 && this.state.invalidUrl) {
+      this.setState({
+        invalidFilename: false,
+      });
+    }
+    if (revision.length === 0 && this.state.invalidUrl) {
```
domoritz: same condition as above

src/components/header/renderer.tsx
```diff
@@ -157,14 +161,31 @@ class Header extends React.Component<Props, State> {
...
     const responseGistCommits = await gistCommits.json();
     if (revision.length === 0) {
       revision = responseGistCommits[0].version;
+      this.setState({
+        invalidRevision: false,
+      });
+    } else {
+      const revGistCommits = await fetch(`https://api.github.com/gists/${gistId}/${revision}`);
+      this.setState({
+        invalidRevision: !revGistCommits.ok,
```
domoritz: how do you know that the id isn't wrong?

src/components/header/renderer.tsx
```diff
@@ -173,9 +194,24 @@ class Header extends React.Component<Props, State> {
         });
         throw Error();
       }
+      this.setState({
+        invalidFilename: false,
+      });
+    } else {
+      const xfilename = Object.keys(gistData.files).find(f => gistData.files[f].language === 'JSON');
```
domoritz: what is xfilename?

src/components/header/renderer.tsx
ArexJambusarwala: The gist-url part (gistId) is checked before this. So all we need to do is check if revision is correct or not.

src/components/header/renderer.tsx
ArexJambusarwala: xfilename is the string which contains the filename from gistData which has language json.
I am checking if the filename user enters and the returned filename are same or not.

domoritz: Please find a better variable name.

src/components/header/renderer.tsx
ArexJambusarwala: This removes the error message of revision if it is empty and if gist-url is false. If this was removed something like this would occur:
corrected11

domoritz: I still don't get why we can't just write

```
if (state.invalidURL) {
  this.setState({
    invalidFileName: filename.length !== 0,
	invalidRevision: revision.length !== 0
  })
}
```

domoritz:
Thanks for cleaning it up. Did you test it carefully?

ArexJambusarwala:
> Thanks for cleaning it up. Did you test it carefully?

I have tested it carefully. :D

src/components/header/renderer.tsx
```diff
@@ -157,14 +161,27 @@ class Header extends React.Component<Props, State> {
     this.setState({
       invalidUrl: !gistCommits.ok,
     });
+    if (this.state.invalidUrl) {
```
domoritz: Why do we condition this on having a valid url?

src/components/header/renderer.tsx
```diff
@@ -157,14 +161,27 @@ class Header extends React.Component<Props, State> {
...
+      this.setState({
+        invalidFilename: filename.length !== 0,
+        invalidRevision: revision.length !== 0,
```
domoritz: Why is a revision with length invalid?

src/components/header/renderer.tsx
domoritz: Why do we condition on having an invalid URL?

src/components/header/renderer.tsx
```diff
@@ -173,9 +190,24 @@ class Header extends React.Component<Props, State> {
...
+      this.setState({
+        invalidFilename: false,
+      });
+    } else {
+      const gistFilename = Object.keys(gistData.files).find(f => gistData.files[f].language === 'JSON');
+      if (this.state.gist.filename !== gistFilename) {
+        this.setState({
+          invalidFilename: true,
+        });
+        throw Error();
```
domoritz: Please don't throw errors.

src/components/header/renderer.tsx
ArexJambusarwala: If the URL is invalid and length of revision/filename is non zero, should I show the revision/filename error?

src/components/header/renderer.tsx
domoritz: I'm very confused by this code. Can you explain why it work?

src/components/header/renderer.tsx
domoritz: I don't know. What's the goal of this logic?

src/components/header/renderer.tsx
```diff
@@ -159,12 +163,21 @@ class Header extends React.Component<Props, State> {
     });
     const responseGistCommits = await gistCommits.json();
     if (revision.length === 0) {
+      this.setState({
+        invalidFilename: false,
+        invalidRevision: false,
+      });
```
ArexJambusarwala: Considering there is already an error in revision and now revision field is left empty we remove the error message. Same for filename. I could've left out the filename and changed it later but it is faulty for the case when the url is invalid and revision is empty and there was a filename error beforehand.

ArexJambusarwala:
Not showing any error in revision and filename if there is already an error in URL.

domoritz:
Thank you!

PR #241, opened 2019-03-08
Add a feature to display error line number

### LINKED ISSUE #240: Incorrect error line number for the editor
For example, it's line 9, not 382 for the following app.  

image

algomaster99:
Fixes #240 

### Fix
These changes add a feature which will also display the number of line where the error exists.

### Demo
Screenshot from 2019-03-09 19-35-32

### Advantage
User can find the error more easily since the line number can be located.

### DIFF
--- a/src/reducers/index.ts
+++ b/src/reducers/index.ts
@@ -48,23 +48,26 @@ function parseVega(state: State, action: SetVegaExample | UpdateVegaSpec | SetGi
     };
   } catch (e) {
     const code = action.spec;
-    const pattern = /\d+/;
+    const pattern = /(position\s)(\d+)/;
     let errorMessage = e.message;
-    const charPos = errorMessage.match(pattern)[0];
+    let charPos = errorMessage.match(pattern);
 
-    if (!isNaN(charPos)) {
-      let line = 1;
-      let cursorPos = 0;
+    if (charPos !== null) {
+      charPos = charPos[2];
+      if (!isNaN(charPos)) {
+        let line = 1;
+        let cursorPos = 0;
 
-      while (cursorPos < charPos && code.indexOf('\n', cursorPos) < charPos && code.indexOf('\n', cursorPos) > -1) {
-        const newlinePos = code.indexOf('\n', cursorPos);
-        line = line + 1;
-        cursorPos = newlinePos + 1;
+        while (cursorPos < charPos && code.indexOf('\n', cursorPos) < charPos && code.indexOf('\n', cursorPos) > -1) {
+          const newlinePos = code.indexOf('\n', cursorPos);
+          line = line + 1;
+          cursorPos = newlinePos + 1;
+        }
+
+        errorMessage = `${errorMessage} and line ${line}`;
       }
-
-      errorMessage = `${errorMessage} and line ${line}`;
+      console.warn(e);
     }
-    console.warn(e);
 
     extend = {
       ...extend,
--- a/src/reducers/index.ts
+++ b/src/reducers/index.ts
@@ -47,11 +47,28 @@ function parseVega(state: State, action: SetVegaExample | UpdateVegaSpec | SetGi
       vegaSpec: spec,
     };
   } catch (e) {
+    const code = action.spec;
+    const pattern = /\d+/;
+    let errorMessage = e.message;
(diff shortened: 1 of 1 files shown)


algomaster99:
@domoritz Can you please review my pull request and suggest any changes required?

domoritz:
Sweet, thank you! 

Before I merge this, could you check whether the monaco editor has a way to access the error it is producing? Maybe we can just hook into that. Thanks!

algomaster99:
@domoritz I couldn't find a way to extract line number from the monaco editor. Although I found a [function](https://microsoft.github.io/monaco-editor/api/modules/monaco.editor.html#getmodelmarkers) in their API which could have helped.
But this function is not defined for the react integration. That's all, I could conclude. Any suggestions you have which can solve this issue?

src/reducers/index.ts
```diff
@@ -47,11 +47,28 @@ function parseVega(state: State, action: SetVegaExample | UpdateVegaSpec | SetGi
       vegaSpec: spec,
     };
   } catch (e) {
+    const code = action.spec;
+    const pattern = /\d+/;
```
algomaster99: #### Before
```
const pattern = /(end|\d+)/
```
#### Now
```
const pattern = /\d+/
```
I only needed to match numbers. This change required me to force-push the branch.

domoritz:
> But this function is not defined for the react integration.

That could be a bug in the typings. I have publishing permissions for https://github.com/superRaytin/react-monaco-editor so you can send a PR to fix it.

domoritz:
See https://github.com/Microsoft/monaco-editor/issues/30 for example.

algomaster99:
@domoritz 
> That could be a bug in the typings.

This function is anyway not mentioned in the `README.md` so I don't think if it requires a pull request. Although `this.refs.monaco.editor` did throw an error: `TS2339: Property 'editor' does not exist on type 'ReactInstance'. Property 'editor' does not exist on type 'Component<any, {}, any>'`. I'll see to it and submit a PR.

domoritz:
Looks like you can't access the editor property. You can ignore the type error and see whether it works first.

algomaster99:
@domoritz I can't seem to find a way to access errors using `react-monaco-editor`.

domoritz:
What about `monaco-editor`?

algomaster99:
@domoritz Haven't checked that out yet. That's a major step to bring in though. I'll set-up a dummy project and try it out first then let you know what will be better. :smile:

domoritz:
Well, you can access the monaco editor directly in the vega editor. You might have to ignore the typings.

algomaster99:
```
const model = this.refs.editor['editor'];
model.getModelMarkers({});
```
Source: https://stackoverflow.com/questions/43463344/getting-errors-from-monaco-editor

This is how I tried solving this issue but it throws an error that `getModelMarkers` isn't a function.

Although
```
model.getModel()
```
works and according to [docs](https://microsoft.github.io/monaco-editor/api/modules/monaco.editor.html#getmodels), `getModel()` and `getModelMarkers()` are in the same class. Line 858 and 867.
Class file: https://github.com/Microsoft/monaco-editor/blob/ddbd7148e4afd00c60e4262f3dc2fb5f1ce18fe2/monaco.d.ts#L858

domoritz:
We're they added in a particular version?

algomaster99:
The `monaco-editor` version we are using and this api's version is same i.e. 0.16.0

domoritz:
Not sure but maybe the last comment in https://github.com/Microsoft/monaco-editor/issues/906 helps. Can you grab the editor from somewhere else other than the onload callback?

algomaster99:
@domoritz Yes, I can get it from anywhere after the editor has mounted

domoritz:
Does it have getModelMarkers then as described in the issue I linked to?

domoritz:
Okay, let's look into this later then.

algomaster99:
@domoritz I didn't add this feature to vega-lite editor. I'll add it and submit a PR soon.

domoritz:
Oh, I missed that. Thanks!

yijiao19:
I'm wondering how to use this feature in the monaco-editor?
I cannot find any description in the API page [https://microsoft.github.io/monaco-editor/api/index.html](url)

Issue #237, opened 2019-03-05
Enhancement : Add Help and Feedback Section


siddhant1:
Since the use of editor is growing fastly we have to find a way to give newcomers help and cater their issues when they find one.

Tasks
- [ ] Add a section named "Help and Feedback" which will be a dropdown
- [ ] Add references along with links to the dropdown list
- [ ] Add reporting bugs to the list
- [ ] Feedback section which holds link to the slack channel
- [ ] Keyboard SHortcuts

I will be providing and a mockup and working on this 
@domoritz please update if you fell something is missing

siddhant1:
For now we have to close #219 , #228 and #229 since everything is covered in this one

domoritz:
"help and feedback" is a bit long for a title

siddhant1:
What should be the title then?

domoritz:
Try it out and see how much space we have. I said earlier we could just have an icon.

siddhant1:
Ok ! I will try it out

ganeshpatro321:
Hey @siddhant1 Have you started working in this issue?

I had some inputs.

siddhant1:
Yes ,  I am working on it :)

siddhant1:
Hey @ganeshpatro ! Sure you can present give your inputs here , would love to hear them

g3o2:
A `?` should do for that purpose.

siddhant1:
Closing fixed !

PR #242, opened 2019-03-09
Add help and usage modal

### LINKED ISSUE #219: Enhancement : Add a keyboard shortcut (cntrl/cmd+enter) to run in manual mode
Most of the editors have this feature that by pressing cntrl/cmd + enter the editor executes the latest saved code.
We should also add a keyboard shortcut in this editor.

algomaster99:
Fixes #219 

### Enhancement
Add a modal which tells the user about:-
- Usage
- Keyboard shortucts
- Links

Screenshot from 2019-03-10 11-54-51

### Usage
- On PC, the shortcut key to open the window is `Ctrl + ?`.
- On Mac, the shortcut key to open the window is `Cmd + '`.
- Use `Esc` key to exit or close the window.

On Mac, `Cmd + ?` triggered help in browser.

### DIFF
--- a/src/components/header/index.css
+++ b/src/components/header/index.css
@@ -558,17 +558,20 @@
   margin: 0;
 }
 
-li {
+.keyboard-shortcuts li {
   margin-bottom: 1em;
 }
 
 kbd {
-  border: 1px solid #666;
-  border-radius: 4px;
-  padding: 3px 5px;
-  margin: 2px;
-  color: #444;
-  text-decoration: none;
+  background-color: #fafbfc;
+  border: 1px #c6cbd1 solid;
+  border-bottom-color: #959da5;
+  border-radius: 3px;
+  box-shadow: inset 0 -1px 0 #959da5;
+  color: #444d56;
+  line-height: 10px;
+  padding: 2px 5px;
+  vertical-align: middle;
 }
 
 /* Media queries */
--- a/src/components/header/index.css
+++ b/src/components/header/index.css
@@ -562,32 +562,13 @@ li {
   margin-bottom: 1em;
 }
 
-li .tutorial-link {
-  display: inline;
-  padding: 0;
-  font-size: 1em;
-  color: blue;
-}
-
-li .tutorial-link:hover {
-  text-decoration: underline;
-}
-
-.site-link-groups a {
-  display: inline;
-  font-size: 1em;
-  background: white;
-  border: 1px solid #ccc;
+kbd {
+  border: 1px solid #666;
   border-radius: 4px;
-  line-height: 2em;
-  cursor: pointer;
-  user-select: none;
-  margin-right: 1em;
-  padding: 0.2em 0.9em;
-}
-
-.site-link-groups a:hover {
-  border-color: #aaa;
+  padding: 3px 5px;
+  margin: 2px;
+  color: #444;
+  text-decoration: none;
 }
 
 /* Media queries */
@@ -597,10 +578,10 @@ li .tutorial-link:hover {
     display: block;
   }
 
-  /* .header-button,
+  .header-button,
   .header a {
     padding: 0 20px;
-  } */
+  }
 
   .auto-run-wrapper {
     width: 125px;
--- a/src/components/header/index.css
+++ b/src/components/header/index.css
@@ -558,6 +558,38 @@
   margin: 0;
 }
 
+li {
+  margin-bottom: 1em;
+}
+
+li .tutorial-link {
+  display: inline;
+  padding: 0;
+  font-size: 1em;
+  color: blue;
+}
+
+li .tutorial-link:hover {
+  text-decoration: underline;
+}
+
+.site-link-groups a {
+  display: inline;
+  font-size: 1em;
+  background: white;
+  border: 1px solid #ccc;
+  border-radius: 4px;
+  line-height: 2em;
+  cursor: pointer;
+  user-select: none;
+  margin-right: 1em;
+  padding: 0.2em 0.9em;
+}
+
+.site-link-groups a:hover {
+  border-color: #aaa;
+}
+
 /* Media queries */
 @media only screen and (min-width: 1200px) {
   .header-button svg,
@@ -565,10 +597,10 @@
     display: block;
   }
 
-  .header-button,
+  /* .header-button,
   .header a {
     padding: 0 20px;
-  }
+  } */
 
   .auto-run-wrapper {
     width: 125px;
--- a/src/components/header/renderer.tsx
+++ b/src/components/header/renderer.tsx
@@ -723,7 +723,7 @@ class Header extends React.Component<Props, State> {
       <div>
         <h2>Help</h2>
         <h4>Keyboard Shortcuts</h4>
-        <ul>
+        <ul className="keyboard-shortcuts">
           <li>
             <kbd>Ctrl</kbd> + <kbd>b</kbd> / <kbd>&#8984;</kbd> + <kbd>b</kbd>: Execute the code in manual mode
           </li>
--- a/src/components/header/renderer.tsx
+++ b/src/components/header/renderer.tsx
@@ -19,7 +19,7 @@ import {
   Trash2,
   X,
 } from 'react-feather';
-import { PortalWithState } from 'react-portal';
+import { Portal, PortalWithState } from 'react-portal';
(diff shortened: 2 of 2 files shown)


algomaster99:
@domoritz Can you please review my pull request and suggest if there are any changes I can make to improve this enhancement?

siddhant1:
Thank you for working on this , we have another ticket opened for this one (#237 ).
If @domoritz likes this design I will extend it to add more features .

src/components/header/index.css
```diff
@@ -558,17 +558,49 @@
...
+
 /* Media queries */
 @media only screen and (min-width: 1200px) {
   .header-button svg,
   .docs-link svg {
     display: block;
   }
 
-  .header-button,
+  /* .header-button,
```
domoritz: Delete if not needed.

src/components/header/renderer.tsx
```diff
@@ -141,6 +143,38 @@ class Header extends React.Component<Props, State> {
...
+    // Handle key press in Mac
+    if (event.keyCode === 222 && event.metaKey && !event.shiftKey) {
+      // 222: Apostrophe(') key
+      this.setState({
+        helpModalOpen: true,
+      });
+    }
+
+    // Handle key press in PC
+    if (event.keyCode === 191 && event.ctrlKey && event.shiftKey) {
```
domoritz: Combine with above

src/components/header/renderer.tsx
```diff
@@ -141,6 +143,38 @@ class Header extends React.Component<Props, State> {
...
+    }
+  }
+
+  public handleHelpModalCloseClick() {
+    this.setState({
+      helpModalOpen: false,
+    });
+  }
+
+  public handleHelpModalCloseEsc(event) {
```
domoritz: This functionality is built into the modal. Please use that.

src/components/header/renderer.tsx
```diff
@@ -687,6 +726,44 @@ class Header extends React.Component<Props, State> {
       </div>
     );
 
+    const helpModal = (
```
domoritz: Please use a portal like any other modal.

src/components/header/renderer.tsx
```diff
@@ -687,6 +726,44 @@ class Header extends React.Component<Props, State> {
...
+            <a className="tutorial-link" href="https://vega.github.io/vega/tutorials/" target="_blank">
+              tutorial
+            </a>{' '}
+            to get familiar with vega components.
+          </li>
+        </ul>
+        <h4>Keyboard Shortcuts</h4>
+        <ul>
+          <li>
+            <strong>Ctrl + b / Cmd + b :</strong> Execute the code in manual mode
```
domoritz: Use kbd tag.

src/components/header/renderer.tsx
```diff
@@ -687,6 +726,44 @@ class Header extends React.Component<Props, State> {
       </div>
     );
 
+    const helpModal = (
+      <div>
+        <h2>Help</h2>
+        <h4>Usage Instructions</h4>
+        <ul>
+          <li>
```
domoritz: I don't find these instructions very helpful. Please remove.

src/components/header/renderer.tsx
```diff
@@ -687,6 +726,44 @@ class Header extends React.Component<Props, State> {
...
+        <h4>Keyboard Shortcuts</h4>
+        <ul>
+          <li>
+            <strong>Ctrl + b / Cmd + b :</strong> Execute the code in manual mode
+          </li>
+          <li>
+            <strong>Ctrl + ? / Cmd + ' :</strong> Open the help window
+          </li>
+        </ul>
+        <h4>Helpful Links</h4>
```
domoritz: Again, not helpful as we already have it.

domoritz:
@siddhant1 This is a start but I left a number of comments above.

src/components/header/renderer.tsx
algomaster99: @domoritz I couldn't figure out how can I add custom keyboard events to toggle Portal that's why I wrote a custom one. I'll see if I can use `<Portal />` and implement this.

domoritz: Portal has a property to close on escape. We already use portals in the editor.

src/components/header/index.css
algomaster99: @domoritz This was interfering with other `a` tags that's why I removed it. If padding is required I can add it.
Right now, it looks like this
Screenshot from 2019-03-10 15-18-54

domoritz: How was it interfering? I think the padding makes it look better.

src/components/header/renderer.tsx
algomaster99: @domoritz Unlike the other modals, this modal is being triggered by a shortcut key. Rest of them were triggered using click only. `onClick` prop would simply trigger the portal. I can't pass a prop which will get triggered when I press `Ctrl + ?`.

domoritz: There must be a way. https://codepen.io/nagyadam2092/pen/eGRxQv is an example that uses setState to open the modal.

algomaster99: I have done the same thing. Even I used `setState` to open the modal. Plus in this example, it doesn't close when pressed `Esc`. So I think this will require writing another function.

src/components/header/index.css
algomaster99: @domoritz It added `padding: 0 20px` to every `a` tag.

src/components/header/renderer.tsx
domoritz: Maybe. I don't have the bandwidth to look deeper into this right now.

src/components/header/index.css
algomaster99: I added it again though. It was just interfering with tutorial link which you asked me to remove.

src/components/header/renderer.tsx
algomaster99: Understandable :smile: . I'll try using `<Portal />` once and see if it's possible.

src/components/header/renderer.tsx
```diff
@@ -875,17 +845,19 @@ class Header extends React.Component<Props, State> {
           </PortalWithState>
 
           {this.state.helpModalOpen && (
-            <div className="modal-background" onClick={this.handleHelpModalCloseClick.bind(this)}>
-              <div className="modal modal-top" onClick={e => e.stopPropagation()}>
-                <div className="modal-header">
-                  <button className="close-button" onClick={this.handleHelpModalCloseClick.bind(this)}>
-                    <X />
-                  </button>
+            <Portal>
```
algomaster99: @domoritz Since I can't use `<PortalWithState />` because of custom events, I used `<Portal />`. `<Portal />` doesn't have `closeOnEsc` prop. Therefore, `handleHelpModalCloseEsc` is required.

algomaster99:
@domoritz I have made the changes accordingly.
modal

src/components/header/renderer.tsx
domoritz: Why do you need custom events?

src/components/header/index.css
```diff
@@ -558,6 +558,19 @@
   margin: 0;
 }
 
+li {
```
domoritz: This is for all li? That's not good to do here.

src/components/header/index.css
```diff
@@ -558,6 +558,19 @@
...
+  margin-bottom: 1em;
+}
+
+kbd {
```
domoritz: Let's use a style that looks more like what GitHub does. I like that one a lot.

src/components/header/index.css
algomaster99: That just gives spaces between the list items. Shall I use `line-height` instead?

domoritz: The problem is that you're setting the css for ALL list elements, no?

algomaster99: Oh okay. I'll make sure it sets the styles for these particular list item only.

src/components/header/renderer.tsx
algomaster99: To add <kbd>Ctrl</kbd> + <kbd>?</kbd> as a shortcut. React only has `onClick`, `onKeyPress` and so on.

algomaster99:
@domoritz Please see the changes. Thank you!
Screenshot from 2019-03-11 09-12-41

Issue #176, opened 2018-10-26
Editor does not set mode from URL


domoritz:
[...]

PR #243, opened 2019-03-09
Mode selection done correctly for every url

ganeshpatro321:
This PR will solve all those mode selection issues for every URL. (Ex: #176 )
As we have either of vega or vega-lite as our mode anytime, so it's safe to have the mode selected and accordingly render state.

@domoritz Please Review.

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -55,6 +55,9 @@ class App extends React.Component<Props & { match: any; location: any; showExamp
     );
 
     const parameter = this.props.match.params;
+    if (parameter.mode) {
+      this.props.setModeOnly(parameter.mode);
+    }
     this.setSpecInUrl(parameter);
   }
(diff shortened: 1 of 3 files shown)


src/components/app.tsx
```diff
@@ -55,6 +55,9 @@ class App extends React.Component<Props & { match: any; location: any; showExamp
     );
 
     const parameter = this.props.match.params;
+    if (parameter.mode) {
+      this.props.setModeOnly(parameter.mode);
+    }
     this.setSpecInUrl(parameter);
```
domoritz: Doesn't this call `setState` twice? Is that a problem?

ganeshpatro321: The first one will just update the mode, you can check the reducer, it won't cause a problem.

domoritz:
Thank you so much!

Issue #247, opened 2019-03-09
Remember location/section when clicking the examples of previous time


mattijn:
I see lots of work on the editor. That's great. So let me add one from a users perspective.

When I click on the example button I hope it can remember the place where I previously clicked one of the examples. Everytime I need to scroll a long way down and search for the section of examples where I was looking right before.

domoritz:
Just as an idea for now. Can you have one window with the examples open and then use CMD+click to open the examples in a new window?

Still, this is a good idea.

ganeshpatro321:
I can work in this issue, assign it to me @domoritz if it is not been picked up by anyone else.

domoritz:
@ganeshpatro321 I sent you an invite (welcome to the team!). You can assign it to yourself now.

ganeshpatro321:
Thank you @domoritz

ganeshpatro321:
Closed via #249

PR #249, opened 2019-03-10
Scroll Position of example portal restored.

ganeshpatro321:
This PR will solve issue #247 

Steps to reproduce:
 - Open example portal
 - Scroll down to any position and select an example to view.
 - Close and reopen the portal, you will see that the scroll position is restored.

I am also attaching a screencast for reference:
record

### DIFF
--- a/src/components/header/renderer.tsx
+++ b/src/components/header/renderer.tsx
@@ -3,6 +3,7 @@ import './index.css';
 import LZString from 'lz-string';
 import * as React from 'react';
 import Clipboard from 'react-clipboard.js';
+import ReactDOM from 'react-dom';
 import {
   Book,
   Code,
@@ -54,6 +55,7 @@ interface State {
   };
   invalidUrl: boolean;
   showVega: boolean;
+  scrollPosition: number;
 }
 
 const formatExampleName = (name: string) => {
@@ -65,7 +67,7 @@ const formatExampleName = (name: string) => {
 
 class Header extends React.Component<Props, State> {
   private refGistForm: HTMLFormElement;
-
+  private examplePortal = React.createRef<HTMLDivElement>();
   constructor(props) {
     super(props);
     this.state = {
@@ -80,6 +82,7 @@ class Header extends React.Component<Props, State> {
         url: '',
       },
       invalidUrl: false,
+      scrollPosition: 0,
       showVega: props.mode === Mode.Vega,
     };
   }
@@ -741,7 +744,18 @@ class Header extends React.Component<Props, State> {
           </PortalWithState>
         </section>
         <section className="right-section">
-          <PortalWithState closeOnEsc defaultOpen={this.props.showExample}>
+          <PortalWithState
+            closeOnEsc
+            defaultOpen={this.props.showExample}
+            onOpen={() => {
+              const node = ReactDOM.findDOMNode(this.examplePortal.current);
+              node.addEventListener('scroll', () => {
+                this.setState({
+                  scrollPosition: node.scrollTop,
+                });
+              });
+            }}
+          >
             {({ openPortal, closePortal, isOpen, portal }) => [
               <span key="0" onClick={openPortal}>
                 {examplesButton}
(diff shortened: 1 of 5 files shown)


ganeshpatro321:
@domoritz Please review.

domoritz:
Looks good. What happens when you change modes after you saved a scroll location?

ganeshpatro321:
@domoritz 

No matter, what mode is selected. It goes to the position where the user have selected an example.

ganeshpatro321:
Look at this recording for better understanding:
record1

domoritz:
What if you save a position for Vega, then the use switches to Vega-Lite and looks at the examples? Don't we show the wrong scroll location?

ganeshpatro321:
Yes, it goes to the scrolled location, do you want to me to change the code to scroll to the top incase of mode change?
@domoritz

ganeshpatro321:
Updated the requirrement @domoritz

src/components/header/renderer.tsx
```diff
@@ -753,13 +773,21 @@ class Header extends React.Component<Props, State> {
                       <div className="button-groups">
                         <button
                           className={this.state.showVega ? 'selected' : ''}
-                          onClick={() => this.setState({ showVega: true })}
+                          onClick={() => {
+                            this.setState({ showVega: true });
+                            const node = ReactDOM.findDOMNode(this.examplePortal.current);
```
domoritz: Can you instead update the redux store here? It seems hacky that you access the Dom.

ganeshpatro321: I dint get you exactly.

Are you saying to call the setScrollFunction() here?
@domoritz

domoritz: Just set the redux state. 

Not sure why we need the scroll position in redux tbh.

ganeshpatro321: I thought, storing it in redux will be safe, if there is a situation thats makes the component to rerender/reload then the local react state will be lost. 
What do you think?

domoritz: We store the redux state in localstorage. I don't know how often we care about preserving the scroll location. 

Just a thought. I always prefer simple solution.

ganeshpatro321: According to me, setting it in the redux store would be fine, because the user will always want to go to the location where he selected the last example.
And if the redux state is stored in the local storage it will make things prettier, because if the user reloads the site, we will have the scroll position and can restore it.

ganeshpatro321: > Can you instead update the redux store here? It seems hacky that you access the Dom.

I fear we cannot do that, because the scroll is restored once the portal is opened, and we need to access the DOM to get the scroll back to top.

domoritz: Why not just reset the scoll location  in redux when the user switches from Vega to Vega-Lite and vice versa?

ganeshpatro321: We can update the redux store when the user switches.
But the location is fetched from the store only when the portal is opened.
You can see th code, there
the scrollTo is called in onOpen of the react-portal. So updating the redux store here won't make the DOM to scroll up.

As the onOpen function is called only once whenever the portal is opened, and toggling between modes won't make the modal reopen which which not trigger the onOpen function.

src/components/header/renderer.tsx
```diff
@@ -741,7 +746,22 @@ class Header extends React.Component<Props, State> {
...
         </section>
         <section className="right-section">
-          <PortalWithState closeOnEsc defaultOpen={this.props.showExample}>
+          <PortalWithState
+            closeOnEsc
+            defaultOpen={this.props.showExample}
+            onOpen={() => {
+              const node = ReactDOM.findDOMNode(this.examplePortal.current);
+              node.scrollTop = this.props.lastPosition;
+              node.addEventListener('scroll', () => {
```
domoritz: Why do we need this if we set the state when we close the modal?

ganeshpatro321: Okay, let me explain you what's happening.

So, when the user opens the portal we are getting the div element in the DOM which shows example as a node.
Then we are scrolling the div to the position which is stored in the redux store.
The above lines are responsible for accessing the DOM and making it scroll to the stored position, the default position is set to 0 in the default state of the redux. So if the user opens the portal for the first time then he stays at the top of the portal.

The setScrollPosition() function which is called onClose of the portal is responsible for storing the location in the redux store.

And when the user reopens the modal, it again fetches the location from the store and scroll to that particular position where the user was there before.

What happens when the user toggles between the modes is, we again access the DOM using the react reference and scroll it to the top.

PR #251, opened 2019-03-10
Mode selection done after verifying

ganeshpatro321:
This is an update to the previous PR #243, it was taking invalid mode and setting the mode as undefined, this PR will make it to verify and then trigger the store action.

@domoritz Please Update it.


domoritz:
Why would the mode be something else? Should we show a warning?

ganeshpatro321:
There can be a situation where the user can change the URL manually.

domoritz:
I don't see that as a big problem. Garbage in, garbage out.

Issue #183, opened 2018-11-29
Why is this broken?


domoritz:
Reported by @kanitw 

Why is this example just `{}`?

[...]

PR #254, opened 2019-03-11
throw an error if URL does not decompress (for issue #183)

bryik:
As discussed in #183, if the portion of the URL containing the LZ encoded Vega specification fails to decompress then `spec` will be `null`. The editor will not update nor will it log any errors.

This PR adds a `throw` when data cannot be decompressed.

### DIFF
--- a/src/components/input-panel/spec-editor/renderer.tsx
+++ b/src/components/input-panel/spec-editor/renderer.tsx
@@ -109,7 +109,7 @@ class Editor extends React.Component<Props, {}> {
     if (spec) {
       this.updateSpec(spec);
     } else {
-      throw new Error(`failed to decompress URL. Expected a specification, but received ${spec}`);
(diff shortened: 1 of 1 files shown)


src/components/input-panel/spec-editor/renderer.tsx
```diff
@@ -108,6 +108,8 @@ class Editor extends React.Component<Props, {}> {
     const spec = LZString.decompressFromEncodedURIComponent(this.props.match.params.compressed);
     if (spec) {
       this.updateSpec(spec);
+    } else {
+      throw new Error(`failed to decompress URL. Expected a specification, but received ${spec}`);
```
domoritz:
```suggestion
      throw new Error(`Failed to decompress URL. Expected a specification, but received ${spec}`);
```

bryik:
I marked this PR as a 'draft' as I haven't managed to get the editor running locally yet (no way to test).

Issue #256, opened 2019-03-11
UX enhancements for gist portal.


ganeshpatro321:
Requirements: 
 - Better input allignments.
 - Better error renderings.
 - Scenario1: Whenever the user clicks on the "Load" button, there is no way that the user can know that a fetch request from gist is been triggered now.
 - Scenario1 Soln: Figure out a way to let the user know that something is loading.

ganeshpatro321:
Closed via #257

PR #257, opened 2019-03-11
UX and UI enhancements for gist portal

ganeshpatro321:
This PR will solve issue #256 .

- What will happen now is, when the user clicks on the load button if the await function for loading gist is called, then the text will change to "Loading..".
- UI of the portal updated.

### DIFF
--- a/src/components/header/index.css
+++ b/src/components/header/index.css
@@ -249,8 +249,6 @@
 }
 
 .gist-optional-input-container {
-  /* float: left; */
-  /* width: 50%; */
   box-sizing: border-box;
 }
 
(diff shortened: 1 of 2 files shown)


src/components/header/index.css
```diff
@@ -249,8 +249,8 @@
 }
 
 .gist-optional-input-container {
-  float: left;
-  width: 50%;
+  /* float: left; */
```
domoritz: remove if not necessary

ganeshpatro321: Actually I wanted to ask if you want to keep the filename and revision inputs in a row or like this:
Screenshot from 2019-03-12 14-35-48

ganeshpatro321:
Updated the PR @domoritz

Issue #255, opened 2019-03-11
docker-compose fails (node-pre-gyp)


bryik:
I'm trying to get the editor running locally with Docker on Windows. 

Everything is fine up until the "Linking dependencies" stage:

```
[3/4] Linking dependencies...
warning " > monaco-editor-webpack-plugin@1.7.0" has incorrect peer dependency "monaco-editor@^0.15.1".
[4/4] Building fresh packages...
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
error /usr/src/app/node_modules/canvas: Command failed.
Exit code: 1
Command: node-pre-gyp install --fallback-to-build
Arguments:
Directory: /usr/src/app/node_modules/canvas
Output:
node-pre-gyp info it worked if it ends with ok
node-pre-gyp info using node-pre-gyp@0.11.0
node-pre-gyp info using node@11.11.0 | linux | x64
node-pre-gyp WARN Using needle for node-pre-gyp https download
node-pre-gyp info check checked for "/usr/src/app/node_modules/canvas/build/Release/canvas.node" (not found)
node-pre-gyp http GET https://github.com/node-gfx/node-canvas-prebuilt/releases/download/v2.3.1/canvas-v2.3.1-node-v67-linux-musl-x64.tar.gz
node-pre-gyp http 404 https://github.com/node-gfx/node-canvas-prebuilt/releases/download/v2.3.1/canvas-v2.3.1-node-v67-linux-musl-x64.tar.gz
node-pre-gyp WARN Tried to download(404): https://github.com/node-gfx/node-canvas-prebuilt/releases/download/v2.3.1/canvas-v2.3.1-node-v67-linux-musl-x64.tar.gz
node-pre-gyp WARN Pre-built binaries not found for canvas@2.3.1 and node@11.11.0 (node-v67 ABI, musl) (falling back to source compile with node-gyp)
node-pre-gyp http 404 status code downloading tarball https://github.com/node-gfx/node-canvas-prebuilt/releases/download/v2.3.1/canvas-v2.3.1-node-v67-linux-musl-x64.tar.gz
gyp info it worked if it ends with ok
gyp info using node-gyp@3.8.0
gyp info using node@11.11.0 | linux | x64
gyp info ok
gyp info it worked if it ends with ok
gyp info using node-gyp@3.8.0
gyp info using node@11.11.0 | linux | x64
gyp ERR! configure error
[...]

PR #261, opened 2019-03-11
Get docker-compose working (fixes issue #255)

bryik:
**Changes:**

- Use Debian as base image as Alpine Linux does not currently work with pre-built canvas binaries

- Update packages before installing rsync (also bash package no longer needs to be installed)

- Remove '/r' characters from vendor.sh

Note that this PR does not eliminate all warnings. 

```bash
warning " > monaco-editor-webpack-plugin@1.7.0" has incorrect peer dependency "monaco-editor@^0.15.1".
```

and 

```bash
npm WARN lifecycle The node binary used for scripts is /tmp/yarn--1552313280836-0.6231136158533865/node but npm is using /usr/local/bin/node itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.
```

still occur, but they do not prevent Docker from working.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,14 @@
-FROM node:11-alpine
+FROM node:11
 
 # Informs Docker that the container listens on the specified port at runtime
 # https://docs.docker.com/engine/reference/builder/#expose
 EXPOSE 8080
 
+# Fetch updated list of packages
+RUN apt-get update
+
 # Install rsync as it is a dependency of ./scripts/vendor.sh
-RUN apk add rsync bash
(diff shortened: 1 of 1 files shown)


bryik:
Not really sure why Travis fails. The logs suggest `yarn lint` is returning an error

```bash
ERROR: /home/travis/build/vega/editor/src/components/header/renderer.tsx:87:7 - The key 'helpModalOpen' is not sorted alphabetically
```

But this PR only changes the dockerfile 🤔

siddhant1:
Looks like a flakey test , let @domoritz restart the test and check

PR #262, opened 2019-03-11
Added cntrl+f11 shortcut for fullscreen

### LINKED ISSUE #122: Keyboard shortcut for full screen mode toggle

siddhant1:
Fixes #122 , will add to keyboard shortcuts roo

### DIFF
--- a/src/components/renderer/renderer.tsx
+++ b/src/components/renderer/renderer.tsx
@@ -1,5 +1,3 @@
-import './index.css';
-
 import * as React from 'react';
 import { Edit3, Maximize } from 'react-feather';
 import { Portal } from 'react-portal';
@@ -9,6 +7,7 @@ import * as vega from 'vega';
 import vegaTooltip from 'vega-tooltip';
 import { Mode, View } from '../../constants';
 import addProjections from '../../utils/addProjections';
+import './index.css';
 
 // Add additional projections
 addProjections(vega.projection);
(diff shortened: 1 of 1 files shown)


domoritz:
Isn't that interfering with the browsers fullscreens shortcut?

siddhant1:
That's why I added cntrl+f11 , it doesn't interfere

siddhant1:
@domoritz can you check if this is working on mac?

src/components/renderer/renderer.tsx
```diff
@@ -113,6 +113,18 @@ class Editor extends React.Component<Props, State> {
   public componentDidMount() {
     this.initView(this.props);
     this.renderVega(this.props);
+    // Add Event Listner to f11 key
```
domoritz: listener

siddhant1: :p Doing

domoritz:
Works. I think it would be good to support CMD + F11 on mac as well.

siddhant1:
> Works. I think it would be good to support CMD + F11 on mac as well.

I will look into it , What is the current behaviour?

siddhant1:
Have you tested it @domoritz ? Tell me if any changes are required

domoritz:
Cmd does not work yet.

siddhant1:
Pushed , can you confirm now?

PR #264, opened 2019-03-11
Fix panel opening

### LINKED ISSUE #263: Opening panels does not work
editor

algomaster99:
Fixes #263 
### Changes
Fixes panel toggle for the first time as well. After clearing local storage.
### Demo
resize

### DIFF
--- a/src/components/input-panel/index.tsx
+++ b/src/components/input-panel/index.tsx
@@ -34,10 +34,9 @@ class InputPanel extends React.Component<Props> {
     }
   }
   public componentDidUpdate() {
-    const compiledVegaPane = this.refs.compiledVegaPane as any;
     if (this.props.mode === Mode.VegaLite) {
-      if (compiledVegaPane.pane2.style.height > LAYOUT.MinPaneSize && !this.props.compiledVegaSpec) {
-        this.props.toggleCompiledVegaSpec();
+      if (this.props.compiledVegaPaneSize === 25) {
+        this.props.setCompiledVegaPaneSize(LAYOUT.MinPaneSize + 25);
       }
     }
   }
--- a/src/components/viz-pane/renderer.tsx
+++ b/src/components/viz-pane/renderer.tsx
@@ -35,9 +35,8 @@ export default class VizPane extends React.Component<Props> {
     }
   }
   public componentDidUpdate() {
-    const debugPane = this.refs.debugPane as any;
-    if (debugPane.pane2.style.height > LAYOUT.MinPaneSize && !this.props.debugPane) {
-      this.props.toggleDebugPane();
+    if (this.props.debugPaneSize === 25) {
+      this.props.setDebugPaneSize(LAYOUT.MinPaneSize + 25);
     }
     if (this.props.error) {
       this.props.showLogs(true);


algomaster99:
@domoritz I have made the changes. Please review :)

domoritz:
Can we make the panel much larger by default (e.g. 30%) but still save the location?

algomaster99:
@domoritz I don't get you exactly. Do you want me to increase the size of the panel when popping it up for the first time? Currently it is 60px.

domoritz:
Yeah. It's tiny right now.

algomaster99:
Okay. So shall I make it 100px?

domoritz:
30% would be good. The debug panel could be smaller.

algomaster99:
30% of full split pane? The size of the compiled vega spec panel and the debug pane.

domoritz:
30% of the height of the vl + compiled spec. The debug page should probably start at like 2000px.

algomaster99:
@domoritz I have made the changes. So for the left split panel, whenever the compiled vega spec pane will be popped up it be pop to 30% of the total height of split pane.
For debug pane, can you please explain?

domoritz:
> whenever the compiled vega spec pane will be popped up it be pop to 30% of the total height of split pane.

Whenever? What if the user opened it before to a particular size? What if the saved size is larger than the size of the window?

> For debug pane, can you please explain?

Just make the default 200 or 300px. It doesn't need to be that large.

algomaster99:
@domoritz Sorry for wrong information. It will pop up to 30% only the first. Resizing will save the height obviously.
For debug panel, I'll set the height to 200px on first pop-up. Is that cool?

domoritz:
Sounds great! I have not tried 200px but that seems like a good number.

src/components/input-panel/index.tsx
```diff
@@ -36,7 +36,7 @@ class InputPanel extends React.Component<Props> {
   public componentDidUpdate() {
     if (this.props.mode === Mode.VegaLite) {
       if (this.props.compiledVegaPaneSize === 25) {
-        this.props.setCompiledVegaPaneSize(LAYOUT.MinPaneSize + 25);
+        this.props.setCompiledVegaPaneSize((window.innerHeight - LAYOUT.HeaderHeight) * 0.3);
```
domoritz: LAYOUT.MinPaneSize is now unused.

src/components/viz-pane/renderer.tsx
```diff
@@ -36,7 +36,7 @@ export default class VizPane extends React.Component<Props> {
   }
   public componentDidUpdate() {
     if (this.props.debugPaneSize === 25) {
-      this.props.setDebugPaneSize(LAYOUT.MinPaneSize + 25);
+      this.props.setDebugPaneSize(200);
```
domoritz: Make this a constant in Layout

src/components/input-panel/index.tsx
algomaster99: `LAYOUT.MinPaneSize` is used some places so can't be removed from `consts.ts`

algomaster99:
@domoritz I have made the changes. Please review :)

src/components/input-panel/index.tsx
domoritz: Shouldn't you use it then to determine the min size?

src/components/input-panel/index.tsx
```diff
@@ -74,8 +74,8 @@ class InputPanel extends React.Component<Props> {
           onDragFinished={() => {
             if (this.props.compiledVegaPaneSize === LAYOUT.MinPaneSize) {
               this.props.setCompiledVegaPaneSize((window.innerHeight - LAYOUT.HeaderHeight) * 0.3);
```
domoritz: use min size here

src/components/input-panel/index.tsx
algomaster99: Initially, there will just header which is 25px in height. Let's say window.innerHeight is 160px(including the top header). And there is the top header which is 60px in heighr
So when clicked for the first time, it will expand to (160-60)*0.3 = 30px height. 

Do you want me to add a condition which will reset it to 25px if the above expression evaluates to less than 25px?

domoritz:
Thank you!

PR #266, opened 2019-03-11
Refactor sharing modal

### LINKED ISSUE #156: Refactor sharing modal
I had another look at the sharing modal and noticed some issues with the code. Instead of changing the UI in `componentDidUpdate` in `exportURL` for example, we should change things in `render`. To clean things up, we should extract all the modals into their own components.

algomaster99:
Fixes #156 
`this.exportUrl()` is no longer called in `componentDidUpdate()`. Instead it is given as callback when `handleCheck` is triggered by toggling the checkbox.

### DIFF
--- a/src/components/header/renderer.tsx
+++ b/src/components/header/renderer.tsx
@@ -82,9 +82,9 @@ class Header extends React.Component<Props, State> {
         type: props.mode,
         url: '',
       },
+      helpModalOpen: false,
       invalidFilename: false,
       invalidRevision: false,
-      helpModalOpen: false,
       invalidUrl: false,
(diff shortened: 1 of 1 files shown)


algomaster99:
@domoritz Any more changes required?

domoritz:
Could you extract the modal in a separate component or is that too much work?

algomaster99:
Will do :)

domoritz:
Let's do that in a follow up PR. I'll merge this.

algomaster99:
Thanks! Shall I do it for all the modals?

domoritz:
probably, yes

Issue #110, opened 2018-06-14
Add visual feedback for format keyboard shortcut


ydlamba:
Same click button effect like we have done for parsing?

ydlamba:
After the new top bar, we need to add visual feedback for parsing/formatting again.

domoritz:
👍

algomaster99:
@domoritz Can you please explain what this issue is about? :) I could work on this probably.

domoritz:
When you format the code for example with a keyboard shortcut, there is no visual indicator. Same for running a spec btw.

siddhant1:
I thinkthis can also be closed @domoritz

PR #269, opened 2019-03-12
Added Acknowledgement for alt+shift+f (format)

siddhant1:
ref #110 

ezgif-4-03f3da69be05

### DIFF
--- a/src/components/header/index.css
+++ b/src/components/header/index.css
@@ -34,10 +34,7 @@
 }
 
 .header-button:hover,
-.docs-link:hover {
-  background: rgba(0, 0, 0, 0.08);
-}
-
+.docs-link:hover,
 .dark-background {
   background: rgba(0, 0, 0, 0.08);
 }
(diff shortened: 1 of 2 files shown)


src/components/header/index.css
```diff
@@ -38,6 +38,10 @@
   background: rgba(0, 0, 0, 0.08);
 }
 
+.dark-background {
```
domoritz: Can we use the right css pseudo class instead?

siddhant1: can you tell me which pseudo class should I include?
Should I hover and click and unhover ?

domoritz: Maybe that the best solution. We could use some css to make the change stay for a few milliseconds.

siddhant1: I think we cannot apply hover effect programatically (i tried) , so instead of making an extra class (.dark-background), should i apply styles using js?

domoritz: Nah, an extra class is okay. How about we share the definition for a pseudo class click as well as the fake click?

siddhant1:
Sounds good

siddhant1:
@domoritz any other changes required?

Issue #267, opened 2019-03-12
UI/UX fixes in help modal


algomaster99:
### Current UI
![Screenshot from 2019-03-12 16-31-39](https://user-images.githubusercontent.com/35191225/54195537-f549d680-44e4-11e9-9254-1b4e333b0486.png)

### Suggested fixes
- Unnecessary scrollbar
- Alignment of icon and text in feedback buttons
- Tips can be written with better alignment
- Spacing between characters and line-height can be more consistent

nandinirai:
@domoritz I'll fix it and send a PR asap

siddhant1:
I was looking at it in the night and dom pushed it with new changes , I have fixes ready for nearly all of them , will push now

siddhant1:
Closing fixed

PR #260, opened 2019-03-11
Extended Help feedback

siddhant1:
Added more features as of #237

PR #270, opened 2019-03-12
Fixed styles (Addresses #267)

siddhant1:
ref #267

### DIFF
--- a/src/components/help-modal/index.css
+++ b/src/components/help-modal/index.css
@@ -9,10 +9,6 @@
   margin: 4px;
 }
 
-.modal-header {
-  margin-bottom: 9px;
-}
-
 .keyboard-shortcuts li {
   margin-bottom: 1em;
 }
--- a/src/components/help-modal/index.css
+++ b/src/components/help-modal/index.css
@@ -65,4 +65,10 @@ kbd {
 .doc-image {
   height: 20px;
   margin: 0px 6px 0px 2px;
+  vertical-align: sub;
+}
+
+.tips a {
+  display: inline-block;
+  margin: 0px;
 }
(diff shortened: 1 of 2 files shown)


siddhant1:
image

@algomaster99 would you like to see any more changes ?

siddhant1:
@domoritz I think we should change the color of the link or do something like this to let people know that this is a link

image

siddhant1:
I have changes ready , tell me if you want this .

algomaster99:
@siddhant1 
- Margin between heading anf buttons can be more consistent
- Font-size of headers should be uniform
These are my suggestions. If you want to incorporate more, you may :)

siddhant1:
Hm , I think I should change them
Thanks for suggestions :)

siddhant1:
Final Styles , also fixed link to docs styles
image

image

domoritz:
I'm not a huge fan of the links as buttons. Can you instead just make them blue with an underline?

src/components/help-modal/index.css
```diff
@@ -65,4 +73,20 @@ kbd {
 .doc-image {
   height: 20px;
   margin: 0px 6px 0px 2px;
+  vertical-align: sub;
+}
+
+.tips {
+  text-decoration: none;
+  color: black;
```
domoritz: Why override it?

siddhant1: Was overriding to make the link a button , removed

siddhant1:
image
Looks like this now

siddhant1:
Removed redundant margin 

Looks like this now 
image

domoritz:
The alignment is still terrible. Also, could you make the follow and reference links actual links instead of buttons?

siddhant1:
Sure

siddhant1:
Do you want something similar to https://github.com/vega/editor/pull/260#issuecomment-471585781

domoritz:
Yes

siddhant1:
image

PR #275, opened 2019-03-13
Display error line number in vega-lite editor

algomaster99:
#### Error line is displayed for vega-lite editor.
Screenshot from 2019-03-13 14-39-33

### Browser compatibility
|                      |Google Chrome|Mozilla Firefox|Safari|
| ----------------- | :------------------: | :----------------: | :----: |
| Line Number | :heavy_check_mark: | :heavy_check_mark: | :x: |

It is not displayed in Safari since, the error message displayed doesn't contain the position of the error and just shows the type of error. So I don't think it's possible to fetch the exact line of error in case of Safari.
53006183_2264450687151693_6592227800915116032_n


domoritz:
What do we show in safari then?

PR #276, opened 2019-03-13
Updated data viewer on signals

### LINKED ISSUE #173: Use the data change API to update the data viewer

ganeshpatro321:
Fixes #173 

The data viewer is updated if any signal is triggered in the chart.

### DIFF
--- a/src/components/data-viewer/renderer.tsx
+++ b/src/components/data-viewer/renderer.tsx
@@ -66,6 +66,10 @@ export default class DataViewer extends React.Component<Props, State> {
 
     const data = this.props.view.data(selected) || [];
 
+    this.props.view.addDataListener(selected, () => {
+      this.forceUpdate();
+    });
+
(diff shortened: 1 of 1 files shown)


ganeshpatro321:
ScreenCast for reference:
dataUpdate

ganeshpatro321:
Shall we remove the refresh button? or keep it for better user experience, if anything wrong happens, or data doesn't come then he can use to refresh the data.
@domoritz

domoritz:
Let's r remove the button. 

Great work!

Just to check. When the data view panel is closed, we don't listen to data changes, right?

ganeshpatro321:
We do listen, as of now, do you want to remove when the panel is closed?

domoritz:
Let's try it for now and see what happens.

domoritz:
Can you remove the refresh button in a follow up?

ganeshpatro321:
Yes, I will look into it and send a PR asap.

PR #277, opened 2019-03-13
Added Warning count in log tab (fixes #147)

### LINKED ISSUE #147: Show warning count in logger tab header
Otherwise it's impossible for users to know that there are warnings when they use the data viewer.

siddhant1:
ref #147 
Capture

### DIFF
--- a/src/components/viz-pane/debug-pane-header/index.tsx
+++ b/src/components/viz-pane/debug-pane-header/index.tsx
@@ -8,6 +8,7 @@ import * as EditorActions from '../../../actions/editor';
 interface Props {
   debugPane?: boolean;
   logs?: boolean;
+  warningsLogger: any[];
 
   showLogs: (val: any) => void;
   toggleDebugPane: () => void;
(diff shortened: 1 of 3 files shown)


domoritz:
Sweet. Ist ist enough to put the count in the store?

siddhant1:
> Sweet. Ist ist enough to put the count in the store?

Didn't get it , Can you explain? redux store?

domoritz:
Sorry, I meant props. Instead of retuning the logs in the end, you could just return the count, no?

siddhant1:
Yes , I think I can make another entry in the reducer , will check and confirm if that gives a performance boost

domoritz:
It probably doesn't matter but I like to keep interfaces lean.

siddhant1:
Ok!

siddhant1:
image
I think we can't access the warns property, so i can't put it inside a war named logsCount

domoritz:
We should probably also show errors and.makenit red when there is an error (or even automatically open the error panel).

siddhant1:
Sounds good !

algomaster99:
@domoritz Accessing length of `warn` isn't possible since it is a private variable in `logger.ts`. I think current solution is the best possible way.

domoritz:
I don't get it. We already run `({(this.props.warningsLogger as any).warns.length})` so why can't we use this earlier?

siddhant1:
Maybe we have to create a copy of the currlogger and then put the value inside logsCount , or maybe the redux connect is made that way to avoid the typings , or it makes a copy of the object and send us

PR #280, opened 2019-03-14
Updates

ganeshpatro321:
Follow up for #276 

Updates: 
 - Removed refresh button.
 - Checked whether the debug panel is open or not.

Removing the listener on the closing of debugPanel will have to be a complex procedure because the debugPanel is not unmounting on hiding the panel. We will have to add and remove it on toggling the panel.

The scenario now is, the data listener is active once the panel is opened, and wont deactivate after closing the panel.

Please let me know if I will proceed with the above procedure of removing the listener on alternative toggling of panel, for this to happen, we will also need some parameters to be pushed into the redux state/ or if we want to avoid redux usage, then we should update the API, and create a api that would remove all the listeners present for the chart.

According to me, sending data to the redux store would be inefficient and hacky, creating one more API to remove all the listeners would be great.
@domoritz

### DIFF
--- a/src/components/data-viewer/index.css
+++ b/src/components/data-viewer/index.css
@@ -48,6 +48,7 @@
 .pagination-wrapper {
   display: inline-flex;
   vertical-align: middle;
+  margin-left: 5px;
 }
 
 .pagination {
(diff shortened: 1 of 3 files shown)


domoritz:
My main concern is the performance overhead. If the debug panel is closed, we should not be updating the table but maybe it doesn't matter since we are using pagination anyway.

Issue #184, opened 2018-11-30
Can't load editor after deadlock in vega


iliatimofeev:
I’ve got a suggestion to the vega editor:
I bumped into vega deadlock bug https://github.com/vega/vega/issues/1464 while the editor was on auto render mode. As result, I couldn’t load vega editor anymore. Magick links with empty document didn’t help.

I have worked it by cleaning local storage of paused the site.

So I suggest to somehow check on editor load that the last render of stored data was successful and if not switch to manual rendering.

ganeshpatro321:
Hey @domoritz , can you explain how to reproduce this, so that i can get an insight of what is happening. I can work in it.
Thank You.

domoritz:
Just put in an invalid spec and see how the editor renders it again when you load the page again.

ganeshpatro321:
@domoritz Ya, its clearing the editor, shouldn't that be the case? Because there is not input?

domoritz:
What if the spec is valid but too large. We could have a timeout, or a separate url that clears the state before opening the editor. Do you want to add the latter?

ganeshpatro321:
@domoritz 

I was looking into this, and I don't think anything can be done here. 
This is going into deadlock due to vega package.

Please correct me if I am misleading.

domoritz:
We'll, what we can do is have a URL to reset the state before loading the editor. That at least would allow a user to use the editor again.

iliatimofeev:
I expected that if I reload editor with broken spec it will switch to
manual rendering.
 I

вт, 19 мар. 2019 г. в 16:04, Dominik Moritz <notifications@github.com>:

domoritz:
Even then the initial rendering will lock the editor.

ganeshpatro321:
Exactly @domoritz @iliatimofeev

domoritz:
So, the fix is to have a URL to clear or reset the editor state completely before loading anything. Can you add that?

ganeshpatro321:
Do we want to reset the state after certain time of loading? @domoritz

domoritz:
no

Just make `https://vega.github.io/editor/#/reset` work.

ganeshpatro321:
I have made a route which clears the localStorage, but still not able to access the route when the editor goes into deadlock

I also tried avoiding the App component and created  a new component, that too didnt work.

Do you have any idea from where can I access things if the editor goes into deadlock. @domoritz

PR #281, opened 2019-03-14
Added debugging configuration for VScode supporting chrome and firefox

ganeshpatro321:
This PR will set the VSCode debugging configuration supporting chrome and firefox. This will help debug the code in situations where the browser does not get the sources/files of the application.
For ex:, deadlock issue like #184 

@domoritz

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,30 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "firefox",
+      "request": "launch",
+      "reAttach": true,
+      "name": "Launch Firefox",
+      "url": "http://localhost:8080",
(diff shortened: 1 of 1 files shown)


ganeshpatro321:
I think there are duplicate commits, could you please squash them. @domoritz

.vscode/launch.json
```diff
@@ -0,0 +1,30 @@
...
+    },
+    {
+      "type": "chrome",
+      "request": "launch",
+      "name": "Launch Chrome",
+      "url": "http://localhost:8080",
+      "webRoot": "${workspaceFolder}"
+    },
+    {
+      "type": "node",
```
domoritz: Does this actually work?

ganeshpatro321:
Yes, they do work, we need to install two extensions(debugger for chrome, and debugger for Firefox) in vscode. @domoritz

domoritz:
I asked about the node configuration.

.vscode/launch.json
ganeshpatro321: Ohh, I dint check this though it was there in default when I created launch.json. Will check it and update the PR.

ganeshpatro321:
Updated @domoritz

PR #285, opened 2019-03-14
Preserve changes while switching editors

### LINKED ISSUE #12: Combine the editor instances

algomaster99:
Changes are now preserved when editors are switched.
### Demo
change

### Further improvement
It will be an enhancement if we could also restore the undo/redo history stack.

This PR doesn't resolve #12 completely but is definitely a step in a right direction.

### DIFF
--- a/src/actions/editor.ts
+++ b/src/actions/editor.ts
@@ -7,6 +7,7 @@ export const PARSE_SPEC: 'PARSE_SPEC' = 'PARSE_SPEC';
 export const SET_BASEURL: 'SET_BASEURL' = 'SET_BASEURL';
 export const SET_COMPILED_VEGA_PANE_SIZE: 'SET_COMPILED_VEGA_PANE_SIZE' = 'SET_COMPILED_VEGA_PANE_SIZE';
 export const SET_DEBUG_PANE_SIZE: 'SET_DEBUG_PANE_SIZE' = 'SET_DEBUG_PANE_SIZE';
+export const SET_EDITOR_REFERENCE: 'SET_EDITOR_REFERENCE' = 'SET_EDITOR_REFERENCE';
 export const SET_GIST_VEGA_LITE_SPEC: 'SET_GIST_VEGA_LITE_SPEC' = 'SET_GIST_VEGA_LITE_SPEC';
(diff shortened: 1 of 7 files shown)


algomaster99:
@domoritz Could you please review my pull request? Thank you!

domoritz:
I think this adds complexity without really solving the underlying issue. Let's try to combine the instances and see what we get from that.

algomaster99:
@domoritz In one case, a split pane is mounted and in another, just one Monaco editor. A new Monaco editor instance creation is necessary actually.
If we still want this feature, I think we might lose resizing of panes. Since we won't be able to use `<Split Pane />` then.

domoritz:
I disagree. We could for example always have a split pane and just hide a panel and split grabber when we only have one editor.

algomaster99:
You mean we can just change styles on toggling? Making one of them `display: none` or `display: flex` depending upon the requirement. Could be a good idea actually.

domoritz:
I don't like it but it's a possibility. 

Better even: make it work so that react does not re-create the editor. It should be possible somehow, I hope.

algomaster99:
> just hide a panel and split grabber when we only have one editor

@domoritz I don't get this exactly. Can you please explain?

domoritz:
> > just hide a panel and split grabber when we only have one editor

> @domoritz I don't get this exactly. Can you please explain?

What you said earlier. (You mean we can just change styles on toggling? Making one of them display: none or display: flex depending upon the requirement. Could be a good idea actually.)

algomaster99:
@domoritz Okay, I'll go for it. One bad repercussion of this idea could be unnecessary load on the browser.

domoritz:
How so?

algomaster99:
Many editors would run at once right. It's a possibility. Not sure.

algomaster99:
@domoritz I have tried doing this to solve the issue. Please review.
demo

### Things to be fixed in this screencast
- Expand vega edition in vega mode
- Put schema as default editorString.

algomaster99:
@domoritz 
demo
### Things done in this screencast
1. I wrote some specs in vega-lite.
2. Swtiched to vega editor.
3. Pressed <kbd>Ctrl</kbd> + <kbd>z</kbd> till the strings got removed.

domoritz:
This needs some more work. Please open a new PR.

PR #288, opened 2019-03-15
Add tab for signal viewer

### LINKED ISSUE #283: Signal viewer
Next to the "data viewer" it would be useful to have a tab with a "signal viewer". 

Especially when developing Vega spec I have to use the JavaScript Console to check the signals under `VEGA_DEBUG.view.data._signals.{x/y/z}`. If this is possible with a tab next to the "data viewer", that would be great.

algomaster99:
Fixes #283 
### Demo
signal

### DIFF
--- a/src/components/signal-viewer/renderer.tsx
+++ b/src/components/signal-viewer/renderer.tsx
@@ -19,13 +19,13 @@ export default class SignalViewer extends React.Component<Props> {
     return (
       <div className="signal-viewer">
         <table>
-          <tr>
-            <th>Signal</th>
-            <th>Value</th>
-          </tr>
           <tbody>
-            {Object.keys(this.props.view[KEY]).map((signal, id) => (
-              <SignalRow key={id} signal={signal} view={this.props.view} />
+            <tr>
+              <th>Signal</th>
+              <th>Value</th>
+            </tr>
+            {Object.keys(this.props.view[KEY]).map(signal => (
+              <SignalRow key={signal} signal={signal} view={this.props.view} />
             ))}
           </tbody>
         </table>
--- a/src/components/signal-viewer/renderer.tsx
+++ b/src/components/signal-viewer/renderer.tsx
@@ -1,67 +1,34 @@
 import React from 'react';
-import * as vega from 'vega';
 
 import { View } from '../../constants';
-import Table from '../table';
+import SignalRow from '../signal';
 
 import './index.css';
 
 const KEY = '_signals';
-const header = ['Signal', 'Value'];
 
 interface Props {
   view?: View;
-  debugPane: boolean;
 }
 
 export default class SignalViewer extends React.Component<Props> {
   constructor(props) {
     super(props);
   }
-  public componentDidMount() {
-    Object.keys(this.props.view[KEY]).map(signal => {
-      this.props.view.addSignalListener(signal, () => {
-        this.forceUpdate();
-      });
-    });
-  }
-  public componentWillUnmount() {
-    Object.keys(this.props.view[KEY]).map(signal => {
-      this.props.view.removeSignalListener(signal, () => {
-        // Do nothing
-      });
-    });
-  }
-  public getValue(signalKey) {
-    let returnValue = '';
-    const currentValue = this.props.view.signal(signalKey);
-    if (typeof currentValue === 'object') {
-      Object.keys(currentValue).map(value => {
-        returnValue += `${value}: ${currentValue[value]}, `;
-      });
-      return returnValue.slice(0, returnValue.length - 2);
-    }
-    return currentValue;
-  }
-  public getData() {
-    let values = [];
-    Object.keys(this.props.view[KEY]).map((signal, id) => {
-      values = [
-        ...values,
-        {
-          id,
-          [header[0]]: signal,
-          [header[1]]: this.getValue.bind(this)(signal),
-        },
-      ];
-    });
-    return values;
-  }
   public render() {
-    const data = this.getData.bind(this)();
     return (
       <div className="signal-viewer">
-        <Table header={header} data={data} />
+        <table>
+          <tr>
+            <th>Signal</th>
+            <th>Value</th>
+          </tr>
+          <tbody>
+            {Object.keys(this.props.view[KEY]).map((signal, id) => (
+              <SignalRow key={id} signal={signal} view={this.props.view} />
+            ))}
+          </tbody>
+        </table>
       </div>
     );
   }
--- a/src/components/signal-viewer/renderer.tsx
+++ b/src/components/signal-viewer/renderer.tsx
@@ -25,6 +25,13 @@ export default class SignalViewer extends React.Component<Props> {
       });
     });
   }
+  public componentWillUnmount() {
+    Object.keys(this.props.view[key]).map(signal => {
+      this.props.view.removeSignalListener(signal, () => {
+        // Do nothing
+      });
+    });
+  }
   public getValue(signalKey) {
     let returnValue = '';
     const currentValue = this.props.view.signal(signalKey);
--- a/src/components/signal-viewer/renderer.tsx
+++ b/src/components/signal-viewer/renderer.tsx
@@ -6,7 +6,7 @@ import Table from '../table';
 
 import './index.css';
 
-const key = '_signals';
+const KEY = '_signals';
 const header = ['Signal', 'Value'];
 
 interface Props {
@@ -19,14 +19,14 @@ export default class SignalViewer extends React.Component<Props> {
     super(props);
   }
   public componentDidMount() {
-    Object.keys(this.props.view[key]).map(signal => {
+    Object.keys(this.props.view[KEY]).map(signal => {
       this.props.view.addSignalListener(signal, () => {
         this.forceUpdate();
       });
     });
   }
   public componentWillUnmount() {
-    Object.keys(this.props.view[key]).map(signal => {
+    Object.keys(this.props.view[KEY]).map(signal => {
       this.props.view.removeSignalListener(signal, () => {
         // Do nothing
       });
@@ -45,7 +45,7 @@ export default class SignalViewer extends React.Component<Props> {
   }
   public getData() {
     let values = [];
-    Object.keys(this.props.view[key]).map((signal, id) => {
+    Object.keys(this.props.view[KEY]).map((signal, id) => {
       values = [
         ...values,
         {
--- a/src/components/signal-viewer/renderer.tsx
+++ b/src/components/signal-viewer/renderer.tsx
@@ -18,6 +18,13 @@ export default class SignalViewer extends React.Component<Props> {
   constructor(props) {
     super(props);
   }
+  public componentDidMount() {
+    Object.keys(this.props.view[key]).map(signal => {
+      this.props.view.addSignalListener(signal, () => {
+        this.forceUpdate();
+      });
+    });
+  }
   public getValue(signalKey) {
     let returnValue = '';
     const currentValue = this.props.view.signal(signalKey);
--- a/src/components/signal-viewer/renderer.tsx
+++ b/src/components/signal-viewer/renderer.tsx
@@ -20,7 +20,7 @@ export default class SignalViewer extends React.Component<Props> {
   }
   public getValue(signalKey) {
     let returnValue = '';
-    const currentValue = this.props.view[key][signalKey].value;
+    const currentValue = this.props.view.signal(signalKey);
     if (typeof currentValue === 'object') {
       Object.keys(currentValue).map(value => {
         returnValue += `${value}: ${currentValue[value]}, `;
@@ -31,7 +31,6 @@ export default class SignalViewer extends React.Component<Props> {
   }
   public getData() {
     let values = [];
-
     Object.keys(this.props.view[key]).map((signal, id) => {
       values = [
         ...values,
--- a/src/components/signal-viewer/renderer.tsx
+++ b/src/components/signal-viewer/renderer.tsx
@@ -1,8 +1,13 @@
 import React from 'react';
 import * as vega from 'vega';
+
 import { View } from '../../constants';
+import Table from '../table';
 
 import './index.css';
+
+const key = '_signals';
+const header = ['Signal', 'Value'];
 
 interface Props {
   view?: View;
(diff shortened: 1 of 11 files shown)


algomaster99:
@domoritz Can you please review my pull request? Please suggest some changes, especially UI ones.

domoritz:
The main point was that users want to see the signal values. This does not do that yet.

algomaster99:
@domoritz Can you give an example? I saw it was `undefined` for most of the signals.

domoritz:
That's okay. Width is defined in many charts.

algomaster99:
signals
@domoritz I have mapped values to signals as well. Please leave a review.

mattijn:
Awesome you are working on this!
Signals also can be data: how does the signal viewer look with the following spec?

```json
{
  "$schema": "https://vega.github.io/schema/vega/v5.json",
  "width": 400,
  "height": 200,
  "padding": 5,
  "data": [
    {
      "name": "table",
      "values": [
        {"category": "A", "amount": 28},
        {"category": "B", "amount": 55},
        {"category": "C", "amount": 43},
        {"category": "D", "amount": 91},
        {"category": "E", "amount": 81},
        {"category": "F", "amount": 53},
        {"category": "G", "amount": 19},
        {"category": "H", "amount": 87}
      ]
    }
  ],
  "signals": [
    {
      "name": "signal_data",
      "update":"data('table')"
    }    
  ]
}
```

algomaster99:
@mattijn 
example

src/constants/default-state.ts
```diff
@@ -38,9 +39,10 @@ export const DEFAULT_STATE: State = {
   format: false,
   gist: null,
   lastPosition: 0,
-  logs: false,
+  logs: true,
```
domoritz: What does this change do?

src/components/signal-viewer/renderer.tsx
```diff
@@ -0,0 +1,55 @@
...
+  debugPane: boolean;
+}
+
+export default class SignalViewer extends React.Component<Props> {
+  constructor(props) {
+    super(props);
+  }
+  public getValue(signalKey) {
+    let returnValue = '';
+    const currentValue = this.props.view[key][signalKey].value;
```
domoritz: Please use view.signal(NAME) instead

src/constants/default-state.ts
algomaster99: bug
This will initially show logs there.

algomaster99:
@domoritz I have made the changes. Please leave a review.

src/constants/default-state.ts
domoritz: Why is this the right solution?

algomaster99: So when page starts up or refreshes, logs will be highlighted.

algomaster99: Initially, when it was set to `false`, when the page use to start-up it, automatically, data-viewer tab became the active one.

algomaster99: @domoritz I have reverted the default state to `false`. Found another solution to fix the problem. :)

algomaster99:
@domoritz It also works when signals are changed.

domoritz:
No, it does not work when the signals change in an interactive Vega chart.

algomaster99:
@domoritz 
Can you please tell what all signals should change when dragging in [this interactive spec](https://vega.github.io/editor/#/url/vega/N4KABGBEAkDODGALApgWwIaQFxUQFzwAdYsB6UgN2QHN0A6agSz0QFcAjOxge1IRQyUa6SgBY6AK1jcAdpAA04KAHdGAExbYwADgAMuxREgpG1fFoBM+w1ELo1axjOpaArDchr0eTDgDaShCgECFQMuioyFqe3ugA+oTcTniwCoEhkBToADasyKn+wJAAHpYGUACeWgDsdK4AvvJgRaU4AGzlkFXtdG2NzSVaAJwWTV1a2vUAujahkLJaAaGhRXgATqbUyGvRsGvwClCo3I4AZt1QewdjWbn50Wpr6C6N6SuQ65vb0fZqAApJGR4Q6QJywbbAnCQX4A5KQV7LEKrDbULY7KFrNCwoEgzHHKjRPHY4H1N5gKbpUkQGZKSCwUzhbIFMBLJFkyDhSK7fZpRGZHJ5aIAQi8eFYqF5y3mcn8ZKCkGQVCBzMgAAFRfFEslYAlAXgsMdWOC1NxlHIxqxCKKolDReL4bNlkVFchldFVDITcoDdwjchLSDLdbhXaJVTlhTQgi2XzOTaoI9npK5rdBbbvPbHUYFrLEUE5RkXW6oR6vT6-fjkE1S6by8bTeaC0Yg9545BgMUcE4qGs8AAKADkxQQOWQA6a8GyEUIfeKfYAlE1yqoNIh54uwBUuzIe-2BxUR9kxxOp6gZxUF-IDCYzHh1-VIAXw6FIyFo-nYxE2zC9cmMjmWQLYI8yMIsUmiQ16zNPwwLodhWAIWRYDAAAyFCwFg2BEEYU48AAaWQCpUPQoVYJyfDCKmYArAfLMpRbPA2w7bdd0HYd4FHccwDndcmi3MBuwhQcDw4o8uIve9HxA58QlfCB32adk40JLFfzo6VFiApsoDAlVIOQL0ZBgpU8DghC8CQ4iwFIky6CwnCKKItCMNs8iCIqKiaL-PkGLbUMpLzGTqUpUIlBpIxD3uXMYylZSoXY0dvKgPAKkINtsicZB0B2dSnmcNsV00dSAC9tm4LRThycF1JNDAnEWcoAEZ9DksAFOAuY4sqSKko+VL0sy7LeryrZohvcx1JkRh4HjdY8hKsqtDmqt2Vq9B6v8JrdFa8NwqgdBiiiwCVnZbgNldSEoHYbhEIldSevinr1LwaaAGsAGFfRxHBmue5gj2iAARRhYB8GQZpZVApl66gNjUJa1nm9lXXgE54w6qVYfULQMZ8q1Wxxul1m4V70f5O5ogAYiGGmHUuYnSYAdXUTQcCKVN410OhGvqIK3zoowXvgV7mSKXzCdBtYSbJjnonYKdhfhXmBagKd2GQJkce0yBxbZ7SjFORhsmyQnZaheX0EVhSQOhZwAbZ8m0zAXRrbzSBTlkPAAGVGFK02BXjRqLD5qMnxVj5-rJ3WBg9oEfb9h2zbARrRF512jDWjbFJA5t8cYiWGZlgO5YV166aJqWmZZxB-Ypn7eafMkZPa07zu+1XkFw3qHu60Soj+4XPtYdu2j+vB7agAAJZBTHwSGAGoADkAEE-mh9SsfhnBlvUlG0a1m3N4PnOdbzsnJel2uncganaf6CvpeZ1cr857mG+k8OhZFwno6KC-SZfiXS2Zc07hzVhrUW2tf76ygIbY2gDzalwdDA221AJ7s2LjgF2Ks5ix29r7Iuddk7B20iHNqn9I4-zPoTPB8dCFOxTqAgsnhuB1RlNnE+v8H4AMTpgq6SD77-2QE-VmAwk48zIXzZuYUPAYDWN-aKH4pQpTStEDKMh+7smSNsS2L0CQ4EqkyFafJThSwlA7DUDxYi6m1Mgvk0hew0MYBrLeUBSiuwVODfeesbbQJzm4iWfdogJQBmMQ2Ljgl2JPhcP+QSoQiUSmE5x2RXHjHThkIRCCoCU1OKceA8A2htCiTbIRIia68KISQ6STdKSOlxmEL8VifA2OVC04Ez1+q7AqKga6Jt1KmNYYTSx6ZmlajdB4hxl0ijhJSZEjxe81Do2YX4k+rQBg9xKE9WByTUnuJwZUQJiV4lbPdjs6IVR0kRULlkm+uT8mFOKW7OBJsKnXxyXkgpRTLn00rsI6uWSeZh2YYgbgPZj4lMLmUrJDR9ncPoW2TE8NvmnPga8hFBl4SN0RM3Op7IVHfkxJgdS2injwD0fGQx1V2QDPMQMYZCZrFjPAhMs6UzTkRPio80CXjFngvotQnxqzDkT02UEpJHKAnIpiXSOJvdEnbIlWk2FFQLDCrbAk0JUAk7YO1s8m57zckPORc8gA8nYeAzBpXau5g0LFyxpHBXqEAA)?

algomaster99:
@domoritz 
### Updated demo
demo

Any more changes to this?

domoritz:
This will crash the browser after a while if we don't unregister the listeners when we don't need them. See the data viewer issues.

algomaster99:
Okay, I'll remove all listeners in `componentWillUnmount`.

domoritz:
What if the visualization gets rerendered? `componentWillUnmount` is not enough.

algomaster99:
@domoritz Then what will be required? I just removed all listeners in componentWillUnmount.

src/components/signal-viewer/renderer.tsx
```diff
@@ -25,6 +25,13 @@ export default class SignalViewer extends React.Component<Props> {
       });
     });
   }
+  public componentWillUnmount() {
+    Object.keys(this.props.view[key]).map(signal => {
```
domoritz: why do you use `[key]`? `key` is not a variable so don't make it look like one.

algomaster99: @domoritz `this.props.view['_signals']` wasn't allowed because of tslint configuration. So had to assign it to a constant.

domoritz: Then find a better name and make it all caps as is standard.

domoritz:
> @domoritz Then what will be required? I just removed all listeners in componentWillUnmount. I don't get exactly what is the issue even after the listener is removed. Can you please explain?

We need to make sure that we unregister events when the user changes the specification.

algomaster99:
@domoritz I don't think so that will be required since I registered signal listener in `componentDidMount` and on editor change, the component `<SignalViewr />` is not mounting so we don't need to unregister listeners in `handleEditorChange`.

domoritz:
So, does that mean that we are not listening to signal changes when the user makes a new visualization?

algomaster99:
No, we are. `this.props.view` is definitely getting updated but the whole component isn't mounting again so we're safe.

algomaster99:
@domoritz Sorry for so many commits though. Please squash and merge if you're satisfied with changes. :)

src/components/signal-viewer/renderer.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export default class SignalViewer extends React.Component<Props> {
+  constructor(props) {
+    super(props);
+  }
+  public componentDidMount() {
+    Object.keys(this.props.view[KEY]).map(signal => {
+      this.props.view.addSignalListener(signal, () => {
+        this.forceUpdate();
```
domoritz: Hmm, instead of forcing an update of the whole component, we should instead just take the new value (which is being passed to the callback) and update only that one value.

algomaster99: @domoritz The whole component here means only a table. We can't just update one table row out of it. Because changing any row value will call render again anyway.

domoritz: Yes, but react can reuse dom elements easier and we don't have to grab signal values from Vega when they don't change. It seems cleaner to me to only update what has actually changed. 

Another option would be to make a react component that takes the view and the name of a signal and then renders only a single row. What do you think?

algomaster99: Yes, that'll be more efficient actually. Good idea! I'll try to implement it.

algomaster99:
@domoritz Please review. I have made the changes :)

src/components/signal-viewer/renderer.tsx
```diff
@@ -0,0 +1,35 @@
...
+    return (
+      <div className="signal-viewer">
+        <table>
+          <tr>
+            <th>Signal</th>
+            <th>Value</th>
+          </tr>
+          <tbody>
+            {Object.keys(this.props.view[KEY]).map((signal, id) => (
+              <SignalRow key={id} signal={signal} view={this.props.view} />
```
domoritz: Why not use the signal as a key here?

algomaster99: @domoritz Okay, will do it

algomaster99: @domoritz Done.

domoritz:
Thank you @algomaster99!

domoritz:
This needs some more work to render values correctly. See how we render values in the data viewer and extract the functionality. Can you do that?

domoritz:
I'm fixing it because I found a lot of other issues with the code. Please have a look at my code to see what I changed.

algomaster99:
@domoritz 
- Fixing the scrollbar is an issue
- Rendering of values

What else is there? I'll fix it if you haven't started already.

domoritz:
I fixed those in Master. I hope it's good now.

algomaster99:
@domoritz 
bug
Something's still wrong. I guess. I'll fix it.

domoritz:
What's wrong? Those signals may be too long to print.

algomaster99:
`autosize` value is set as `{ type: 'pad' }`. It is showing the the html markup

domoritz:
Ups, yeah, that's wrong. I tried to reuse the code and it looks like made a bug somewhere. Thanks for pointing nit out. Can you send a PR?

algomaster99:
@domoritz 
Screenshot from 2019-03-22 20-19-23
I tried one of the graphs for experimentation. The function [`formatValue`](https://github.com/vega/vega-tooltip/blob/master/src/formatValue.ts) returns different things on the basis of input:-
- An array of objects if an array is give as input 
- A html table when an object is given as input i.e. `autosize` is coming as a table.

And to render html(table in this case), react provides an attribute called `dangerouslySetInnerHTML` which not advised to use since this can make one site prone to XSS.

I suggest we write our own recursive(if required) function to render objects such as `{type: pad}`.

domoritz:
I don't think we want a nested table. Instead, we should just have the json. It should be easy but I must have used the wrong function somewhere.

algomaster99:
@domoritz formatValue function behaves like that only. Suppose you give it `{a:1, b:2}`. It will return a table of two rows. Okay, I'll copy the formatValue function but make sure it doesn't append html tags to it.
One more thing, are you okay with array in there? For example, `highlight_tuple_fields`.

domoritz:
I see. That's not what we want, right. Do we have the same bug in the data viewer?

domoritz:
> One more thing, are you okay with array in there? For example, highlight_tuple_fields.

I don't understand the question.

algomaster99:
@domoritz No, we don't have the same bug there because there we render the whole table at once. In signal viewer, we render a row at a time.
> I don't understand the question.

I was just questioning the aesthetes of `highligh_tuple_fields: [{"field": "_vgsid_","type", "E"}]`. Does it look fine? I could be parsed like this: `(field: _vgsid_, type: E)`.

PR #291, opened 2019-03-15
Browser crashing bug solved

### LINKED ISSUE #282: Data viewer makes the browser crash
I see that https://github.com/vega/editor/issues/173 is merged, but if I try this locally with example https://vega.github.io/editor/#/examples/vega-lite/bar it doesn't perform well. Actually my browser just crashes. 
See the following screencast
Kapture 2019-03-14 at 20 52 06

Not sure if I miss something?

ganeshpatro321:
This PR fixes #282 

Please review @domoritz

### DIFF
--- a/src/components/data-viewer/renderer.tsx
+++ b/src/components/data-viewer/renderer.tsx
@@ -35,7 +35,12 @@ export default class DataViewer extends React.Component<Props, State> {
   }
 
   public handleChange(option) {
-    this.setState({ selectedData: option.value, currentPage: 0 });
+    if (this.state.selections.length !== 0) {
+      this.props.view.removeDataListener(this.state.selections[0], () => {
+        /* */
+      });
+    }
+    this.setState({ selectedData: option.value, currentPage: 0, selections: [] });
   }
 
   public handlePageChange(option) {
--- a/src/components/data-viewer/renderer.tsx
+++ b/src/components/data-viewer/renderer.tsx
@@ -17,6 +17,7 @@ interface Props {
 const initialState = {
   currentPage: 0,
   selectedData: '',
+  selections: [],
 };
 
 type State = Readonly<typeof initialState>;
@@ -34,11 +35,6 @@ export default class DataViewer extends React.Component<Props, State> {
   }
 
   public handleChange(option) {
-    if (this.state.selectedData !== '') {
-      this.props.view.removeDataListener(this.state.selectedData, () => {
-        /* */
-      });
-    }
     this.setState({ selectedData: option.value, currentPage: 0 });
   }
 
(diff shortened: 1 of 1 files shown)


domoritz:
Can you explain it, please?

ganeshpatro321:
Previously the datalisteners were added multiple times for the same data source which caused the page to render multiple times. Now we are checking whether the listener for that particular data source is present or not and then adding the listener.

There was such condition, on 10th or 15th click, the page was rerendering some thousand of times, which caused the web page to crash.

@domoritz

domoritz:
This is not a good fix. We should unregister eventlisteners or ignore events better. It's also not good to use a list to check for duplicates. Better use an object.

ganeshpatro321:
Ohh I see, I will update it.
I tried removing event, in fact, I did, but still I had to check through a list, to get the work done. I will look into it again and update the PR asap. @domoritz

ganeshpatro321:
Removed listeners, please review @domoritz

domoritz:
Wait, so of you together the listener and not remove it from the list, don't we forget to reregister it later?

ganeshpatro321:
No it won't, I also had the same confusion we are clearing the whole list. That won't be a problem @domoritz.

domoritz:
Why do we need a list now? Shouldn't we always have exactly one listener at a time?

ganeshpatro321:
We need a list, because the listener is removed when the data source changes, but on different signal like changing the values we need the list to verify is there any listener. @domoritz

domoritz:
Changing what values?

ganeshpatro321:
Let me share a screencast which will make things clear.

domoritz:
Thanks! I'm a bit confused about all of this and it seems like a hack on top of a hack.

ganeshpatro321:
It is quite simple @domoritz 

dataSourceChange

The above recording shows the change of data source, or change of category, in this case while changing we are removing and adding data listeners.

signal

This recording shows signal triggerings, here the need of list comes in, which checks if there is listener present for that particular data, if yes then no more listener will be added and if no, the listener will be added.

domoritz:
Correct me if I'm doing but I think we only need a single listener at a time. The listener is for the datasets the user is looking at right now.

ganeshpatro321:
No @domoritz

The listener is taking category wise data source as inputs. For ex, the above example has no. Of data sources equal to no of categories present in drop down.

domoritz:
Yeah, but we don't have to listen to data sources that the user has not currently selected. Does that make sense?

ganeshpatro321:
No that doesn't make any sense and we aren't listening to data source that user has not selected.

domoritz:
> The listener is taking category wise data source as inputs. For ex, the above example has no. Of data sources equal to no of categories present in drop down.

I guess I'm not understanding this then. Could you help me understand this? Sorry for being so pedantic but I'd like us to get the design right.

ganeshpatro321:
Sure @domoritz

Give me some time, I think we can avoid using lists and instead use a Boolean. I will try it out and if that works it will be easier to understand.
I will message you asap.

domoritz:
Replaced by https://github.com/vega/editor/pull/302

Issue #297, opened 2019-03-18
Signal tab viewer overflow-y scroll missing


mattijn:
I was checking this new signal viewer tab (great btw!) on a laptop, but noticed there is no `overflow-y: scroll` functionality like there is in the data viewer. If there are many signals the current implementation is little hard to use on smaller screens.

See screencast:
<img src="https://user-images.githubusercontent.com/5186265/54556430-23358c00-49b9-11e9-8815-3eb899a24ace.gif" width="250">

ganeshpatro321:
I will look into it @mattijn

PR #299, opened 2019-03-19
fixed scroll styles (#297)

siddhant1:
ref #297

### DIFF
--- a/src/components/signal-viewer/index.css
+++ b/src/components/signal-viewer/index.css
@@ -1,6 +1,7 @@
 .signal-viewer {
   padding: 10px 20px 15px 20px;
   width: fit-content;
+  max-width: 100%;
   height: 90%;
   font-size: 0.9em;
   overflow: auto;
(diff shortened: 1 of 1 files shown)


src/components/signal-viewer/index.css
```diff
@@ -2,6 +2,7 @@
   padding: 10px 20px 15px 20px;
   width: fit-content;
   max-width: 100%;
+  height: 90%;
```
domoritz: Why is this the right fix?

siddhant1:
Giving the section in which the table is present a height kicks in the
overflow:scroll after that height

domoritz:
Why is it exactly 90%? It seems so arbitrary.

siddhant1:
Tried 100% to take the whole width of the upper wrapper , felt that it should be 10% smaller . I can use pixels but it wouldn't be responsive

domoritz:
You can use 100% and padding, no?

siddhant1:
Sure , let me have a look

siddhant1:
Actually 100% is okay too

PR #301, opened 2019-03-19
Basic layout for writing tests using jest

ganeshpatro321:
Tests can be written in files in the form <filename>.test.js

I have created a dummy test to show the layout in the file app.test.js

you can run "yarn test" to run the test files. 

@domoritz

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,0 +1,3 @@
+module.exports = {
+  presets: ["@babel/preset-env", "@babel/preset-react"]
+};
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,3 +1,0 @@
-module.exports = {
-  presets: ["@babel/preset-env", "@babel/preset-react"]
-};
--- a/src/app.test.js
+++ b/src/app.test.js
@@ -1,0 +1,3 @@
+test("Fake Test", () => {
+  expect(true).toBeTruthy();
+});
--- a/package.json
+++ b/package.json
@@ -4,17 +4,22 @@
   "homepage": "http://vega.github.io/editor",
   "private": true,
   "devDependencies": {
+    "@babel/preset-env": "^7.3.4",
+    "@babel/preset-react": "^7.0.0",
     "@types/node": "^11.10.4",
     "autoprefixer": "^9.4.10",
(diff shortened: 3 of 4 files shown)


babel.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  presets: ["@babel/preset-env", "@babel/preset-react"]
```
domoritz: Just curious, why do we need babel when we have typescript?

src/app.test.js
```diff
@@ -0,0 +1,3 @@
+test("Fake Test", () => {
```
domoritz: Let's write the tests in typescript. See vega-lite for how you could set it up.

siddhant1:
This setup will remove tree-shaking from the app , also configure jest to include local storage since we are using it , also I prefer react-testing-library or enzyme for testing . I have all the setup ready from the previous app , should I make a pr?

domoritz:
My requirements:
* As few dependencies as possible
* Support Typescript
* Support testing the app
* Easy to use. 

If you check those boxes, yes, please send a PR!

siddhant1:
The app can be tested only with jest but that'll be a hassle to maintain , also the following pr will make tree-shaking disabled which will slow down the app. Also no support for local storage too ,  I will have a look to improve this .

siddhant1:
There is a global babel config which will disable tree shaking in general (because it will transpile esm)

ganeshpatro321:
> My requirements:
> 
>     * As few dependencies as possible
> 
>     * Support Typescript
> 
>     * Support testing the app
> 
>     * Easy to use.
> 
> 
> If you check those boxes, yes, please send a PR!

I used jest because it is light weight and easy to use. It provides features like snapshot testing which will make make component testing easy.

I was not aware of the breaking of tree-shaking and searched the web for the same, but couldn't find anything for the same.

@domoritz I will recheck whether we need babel or not for typescript, I followed official jest docs while setting up the configuration and ended up here.

Please let me know if you have any further doubts.
@siddhant1

domoritz:
Okay, since this is a bit more work, let's close this PR. Please open a new one when it's ready.

PR #304, opened 2019-03-21
Fixed some  examples broken

siddhant1:
So this is an interesting bug,
To render the table we are mapping over the Object.keys and before that we are checking if currentValue === object .
This is interesting because we are getting a null value and type of null is actually an Object (this is a js bug duh!) so Object.key(null) is an error . 
An easy fix is to check if the value is not falsy.


domoritz:
Thanks!

siddhant1:
Looks like I broke some lint rules, checking

siddhant1:
I am not sure why the tests are breaking , @domoritz please have a look

domoritz:
Hmm, looks like github is partially down.

PR #309, opened 2019-03-23
Fix signal viewer

algomaster99:
Fixes rendering of values. Modified `formatValueLong` function to our own needs.

### Further improvements
- [x] Fix scrolling
- [x] When spec is edited, signals don't change on further actions since component doesn't mount.
- [ ] Write a better function to render signal values

### DIFF
--- a/src/components/signal-viewer/index.css
+++ b/src/components/signal-viewer/index.css
@@ -1,30 +1,50 @@
 .signal-viewer {
-  margin: 10px;
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  padding: 10px 20px 15px 20px;
+}
+
+.signal-table {
   width: fit-content;
   max-width: 100%;
   font-size: 0.9em;
   overflow: auto;
-  flex: 1;
+  border: 1px solid #dfe2e5;
 }
 
-.signal-viewer table {
+.signal-table table {
   max-width: 100%;
   table-layout: auto;
   border-collapse: collapse;
 }
 
-.signal-viewer th tr:first-child > * {
+.signal-table th tr:first-child > * {
   border-top: 0;
   border-bottom: 2px solid #dfe2e5;
 }
 
-.signal-viewer th {
+.signal-table tbody tr:last-child > * {
+  border-bottom: 0;
+}
+
+.signal-table tbody > * > td:first-child,
+.signal-table th > * > th:first-child {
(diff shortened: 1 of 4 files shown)


siddhant1:
I guess the scrolling was already fixed

domoritz:
Wait, was the code to format values in the data table right but it's wrong for signals?

algomaster99:
@domoritz Basically, when the signal value was an object, it was returning a html table. That's why we needed to re-write the function. For other dataTypes, it was working fine. But i think for nested objects it will create an issue so I re wrote the function. Nested objects example: `[{}, {}]`
@siddhant1 There are some fixes like:-
- Horizontal scroll
- When I scroll to bottom, can't see the last entry
- Color of the scroll thumb

domoritz:
What did you do about "Horizontal scroll"?

algomaster99:
@domoritz horizontal scroll is currently disabled somehow

domoritz:
Okay, I want to avoid it but we should not disallow it. Good observation.

algomaster99:
@domoritz 
Scrolling has been fixed :)
scroll

algomaster99:
[Commit](https://github.com/vega/editor/pull/309/commits/f65ab7cf192c13481f1490300aa2c95416f8978d) fixes second checkbox.
Signal listeners were added in componentDidUpdate because if we leave the signal tab opened, and then edit the spec, the signal listeners were added to `prevProps` which caused dysfunction in real-time change of signal values in the table afterwards. And since the component is already mounted, new signal listeners weren't added. So `componentDidUpdate` just adds the signal listener to a new set of `props`.

domoritz:
Notice that I pushed some style changes yesterday so scrolling might already be fixed in Master.

algomaster99:
@domoritz when I rebased master onto branch, scrolling wasn't fixed. I have fixed it in this PR though. One objective remains now, I'll try to optimize the function I have written. I am thinking of using recursion to format signal values.

domoritz:
I think this PR is mixing too many things now. Can you make separate PRs that fix individual bugs? Thanks!

algomaster99:
@domoritz you can review this PR then. The optimization can be left for another PR.

algomaster99:
I'm closing this. I'll make another pull request for the bugs.

domoritz:
Thanks

PR #311, opened 2019-03-24
Used bindActionCreators

ganeshpatro321:
Refactored

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -4,6 +4,7 @@ import * as React from 'react';
 import { connect } from 'react-redux';
 import { withRouter } from 'react-router-dom';
 import SplitPane from 'react-split-pane';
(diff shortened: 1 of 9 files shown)


algomaster99:
@ganeshpatro321 Thanks for giving your time to this issue. :) But this was a good-first-issue and was intended to be solved by new contributors. You're quite experienced with the code base so I would recommend you to solve some major issues.
But obviously, if a good-first-issue remains stale for some time, one may fix it.

ganeshpatro321:
@algomaster99 

Thats right, we can close this and keep the issue open.
I did this to keep the code maintained and updated!

cheers.

algomaster99:
@ganeshpatro321 no need of closing this. People can use this as a reference to fix the whole code :)

domoritz:
Thank you for the fixes! This will make our typings much safer.

PR #313, opened 2019-03-24
Modify styles for scroll bar to make it uniform

algomaster99:
Modify scroll bar styles and resizing behaviour so it matches with the data viewer


domoritz:
Hmm, can you instead use the same style? I don't like to repeat the same style.

algomaster99:
@domoritz Shall I change revert the colour of the scroll bar? This PR will make the colour same as that of data viewer.

domoritz:
I honestly don't care much about the color of the scrollbars. Let's keep the default.

domoritz:
Can you merge the styles for the signal and draw viewer tables?

algomaster99:
@domoritz I have made the changes, please review.

src/components/signal-viewer/renderer.tsx
```diff
@@ -18,7 +20,7 @@ export default class SignalViewer extends React.Component<Props> {
   public render() {
     return (
       <div className="signal-viewer">
-        <div className="signal-table">
+        <div className="data-table">
```
domoritz: This is not acceptable. Let's find a new name that encompasses the different tables.

algomaster99: @domoritz I have used the same names since I need to apply the same styles. If I change it, merging styles wouldn't be possible.

domoritz: We should have a separate type for `tables`.

algomaster99: I am thinking to name them `debug-table`.

domoritz: Why not just create a style for `table` and not even add a class?

algomaster99:
#### Horizontal scroll behaviour
scroll

src/components/signal-viewer/index.css
```diff
@@ -1,42 +1,6 @@
...
-.signal-viewer td:first-child {
-  word-break: keep-all;
-}
-
-.signal-viewer tr:nth-child(odd) {
-  background-color: #f6f8fa;
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  padding: 10px 20px 15px 20px;
```
domoritz: I think it was better before. Why did you change this?

algomaster99: This was because when we scrolled to bottom, we couldn't see the last entry in the table. This ensured that the even some part of table will never go out of the view.

domoritz: I have not had that problem since my fixes in master.

algomaster99: ### master
Screenshot from 2019-03-25 10-22-22

### scroll-bar-style
Screenshot from 2019-03-25 10-20-58

Notice the right padding.

domoritz: This is not a good enough fix for this issue. See how the tabs are over the table. Please revert your change and fix the issue in a new PR.

algomaster99: I'll combine these three tabs in a hamburger and submit a new PR?
These changes only make the vertical scroll bar visible so that user can know it's a scrollable element.

algomaster99: @domoritz Reverting it will cause some problems. Before this commit was made, the structure of the signal-viewer was like this:-
```
 <div className="signal-viewer">
        <table>
          <tbody>
            <tr>
              <th>Signal</th>
              <th>Value</th>
            </tr>
            {this.getSignals().map(signal => (
              <SignalRow key={signal} signal={signal} view={this.props.view} />
            ))}
          </tbody>
        </table>
      </div>
```
But now it has been changed to
```
 <div className="signal-viewer">
        <div className="editor-table">
          <table>
            <tbody>
              <tr>
                <th>Signal</th>
                <th>Value</th>
              </tr>
              {this.getSignals().map(signal => (
                <SignalRow key={signal} signal={signal} view={this.props.view} />
              ))}
            </tbody>
          </table>
        </div>
      </div>
```

domoritz: Why not just make this

```
<div className="signal-viewer">
          <table className="editor-table">
            <tbody>
              <tr>
                <th>Signal</th>
                <th>Value</th>
              </tr>
              {this.getSignals().map(signal => (
                <SignalRow key={signal} signal={signal} view={this.props.view} />
              ))}
            </tbody>
          </table>
      </div>
```

algomaster99: Good idea! I'll try that remove the unnecessary styles.

algomaster99: @domoritz Seems like the extra div acted as a container from which if the table overflowed, scroll bars would appear.

domoritz: I see. I think the table style itself should not have that role. See how I handled it for the signal table.

algomaster99: The extra div was the place the scrollbars were attached. If I apply `overflow: auto` to the outermost div, since it takes the whole width of the container, the scrollbars will then appear sticking to the the container. The extra div is just a place for holding the scroll bars.

domoritz: That's not true as you can see on master right now. You can handle overflow differently for the signal table and the data viewer.

algomaster99: I copied the same styles and structure for signal and data viewer that's why they both were working.
Plus data viewer also had the the extra div.

domoritz: The data viewer needed it because it has the additional buttons. The signal viewer does not.

algomaster99:
```
<div className="data-table">
    <table>
      <thead>
        <tr>{headerNodes}</tr>
      </thead>
      <tbody>{tableBody}</tbody>
    </table>
</div>
```
I was talking about this div.

domoritz:
Replaced by #14b61e7

PR #314, opened 2019-03-24
Add signal listeners in componentDidUpdate

algomaster99:
Signal listeners were added in componentDidUpdate because if we leave the signal tab opened, and then edit the spec, the signal listeners were added to `prevProps` which caused dysfunction in real-time change of signal values in the table afterwards. And since the component is already mounted, new signal listeners weren't added. So `componentDidUpdate` just adds the signal listener to a new set of `props`.


domoritz:
Does component did update get called after mounting? 

Can you try this while switching examples from Vega to Vega-Lite?

algomaster99:
@domoritz It's never called just after mounting.
While switching examples from vega to vega-lite, since the component is already mounted, `componentDidUpdate` is called.

PR #318, opened 2019-03-25
Reset localStorage

ganeshpatro321:
Created a route "localhost:8080/reset.html", with a button to clear the local storage.
We weren't able to access the app component, so had to create a ".html"  file to reset things.

This can be considered as a WIP pr, we need to update the URLs and may be have automated things i.e., remove button before merging.

@domoritz Please review.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -8,7 +8,7 @@
     />
     <meta name="theme-color" content="#325" />
     <link rel="shortcut icon" href="https://vega.github.io/favicon.ico" />
-    <title>Vega Editors</title>
+    <title>Vega Editor</title>
   </head>
   <body>
     <div id="root">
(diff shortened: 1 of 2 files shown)


ganeshpatro321:
Screencast:

reset

domoritz:
I suggested to just add a route. Why did you decide to make a new age instead?

ganeshpatro321:
We aren't able to access the current age when the browser crashes, so I had to create a new age to clear the localStorage.

Can we access it @domoritz 
Correct me if I am wrong.

algomaster99:
@ganeshpatro321 We can clear the site data using this
Screenshot from 2019-03-25 22-30-40

siddhant1:
I personally suggest moving from localstorage API , but a temporary solution would be to tell the user to clear the localstorage himself rather than another route.

ganeshpatro321:
I appreciate you attention guys,

@algomaster99 that's basic, but we wanted a easier way for the user to clear storage.

@siddhant1 yes, but that won't be a good user experience.

Our idea was to have a route which will clear the localStorage and redirect to the home page of editor.

domoritz:
Replaced by bfcac6b

Issue #317, opened 2019-03-25
Initially opening of the data viewer causes crashing of application


algomaster99:
![bug](https://user-images.githubusercontent.com/35191225/54898833-b576ec00-4ef3-11e9-8995-d00e04c695ec.gif)

ganeshpatro321:
Thanks for reporting, will look into this.

PR #319, opened 2019-03-25
Condition refactored

ganeshpatro321:


ganeshpatro321:
Fixes #317

domoritz:
That doesn’t look like the right fix. Instead, we should use the 2nd entry when the list has more than one item but the first when there Is only one.

ganeshpatro321:
Changes done accordingly. @domoritz

ganeshpatro321:
Ohh, dint see the fix. Closing this

PR #322, opened 2019-03-25
Fix height of debug pane header when resizing vertically

### LINKED ISSUE #321: Debug pane header tabs getting squeezed when resized vertically
### Problem
Screenshot from 2019-03-26 01-54-01

This can be fixed by giving the style `min-height: 25px` to  class `debug-pane-header`.

algomaster99:
Fixes #321 
Screenshot from 2019-03-26 02-09-40


domoritz:
Cannot reproduce the bug.

Issue #316, opened 2019-03-25
Cannot open vega panel


domoritz:
![editor](https://user-images.githubusercontent.com/589034/54889960-5a49e880-4e64-11e9-9b80-01d1693b000a.gif)

algomaster99:
Even I'm facing this bug in Firefox though not in Google Chrome.

siddhant1:
I am facing this bug after the pr , I tried solving it and faced some issues , I will have a look at it again and generate a fix

domoritz:
We don't yet have a clean fix.

PR #324, opened 2019-03-26
[WIP] Fix resizing of compiled spec pane

algomaster99:
Addresses #316 

- [x] Fix the jamming of compiled spec pane header when the pane's size was increased.
- [ ] Fix flickering of the spec editor.

bug

I used a function [`editor.layout()`](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.istandalonecodeeditor.html#layout) to resolve this bug.

Other possible solution could be a minor CSS bug because the spec-editor could easily expand but can't shrink. If anyone has any better solution for this, please provide suggestions.


domoritz:
This does not look like the cleanest solution. Let's try to find a cleaner one that is only css.

algomaster99:
@domoritz Yeah, I'm looking for it.

PR #328, opened 2019-03-28
Modify default editor string

### LINKED ISSUE #139: Loading the editor initially neither activates vega, nor vega-lite mode

algomaster99:
Fixes #139 

Since the default mode is set as Vega-Lite, the editor string should also default to `VEGA_LITE_START_SPEC`.


algomaster99:
@domoritz Can you look into this and suggest some changes if there are?

domoritz:
Thank you

PR #330, opened 2019-03-28
Fix toggling when formatting

### LINKED ISSUE #93: Cannot format spec
screen shot 2018-05-21 at 15 59 00

algomaster99:
Fixes #93 
format


domoritz:
We want to use the more compact formatter. This PR does not fix the problem.

PR #332, opened 2019-03-30
Added a link to Gist in Gist loading modal.

takedakun:
I added a link to Gist in Gist loading modal because typing gist url was a pain.


domoritz:
This just opens the gist website. Can you just make the text 'gist' a link?

domoritz:
It already says gist is the title

PR #334, opened 2019-04-01
Extracted share and export modals

ganeshpatro321:
Towards #156

### DIFF
--- a/src/components/header/export-modal/index.css
+++ b/src/components/header/export-modal/index.css
@@ -1,0 +1,88 @@
+.export-content > p {
+  padding: 30px 0 0 0;
+  margin: 0;
+}
+
+.export-buttons {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.export-button {
+  min-width: 280px;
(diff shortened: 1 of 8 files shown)


domoritz:
@siddhant1 can you help review this?

domoritz:
Hmm, this is just one modal extracted rather than also making the modal itself a shared component. Do you think it makes sense to do that?

ganeshpatro321:
Its better to extract components and keep them inside the header directory itself. It will help for better understanding of the file as well as reducing the header file size.

What do you think @siddhant1 ?

siddhant1:
> @siddhant1 can you help review this?

Sure

> Its better to extract components and keep them inside the header directory itself. It will help for better understanding of the file as well as reducing the header file size.

Makes sense , Since we are following container-presentor approach, we can make `HEADER` component our container and present other components inside it.

domoritz:
@ganeshpatro321 Please share a summary of our discussion on slack so we document it here.

Issue #292, opened 2019-03-16
Back button not redirecting when in full-screen mode


algomaster99:
### Actual
![back](https://user-images.githubusercontent.com/35191225/54478677-f5c2d400-483a-11e9-9167-37cfd86b59e2.gif)

### Expected
Back button should redirect to the edit page.

domoritz:
I think this is an instance of a larger number of problems with the back button. It is generally hard to get navigation to work with state in React.

algomaster99:
@domoritz Will see to it soon. Currently, I'm working on #172

PR #335, opened 2019-04-02
Enable browser back button when full screen mode 

takedakun:
Fixed issue #292 and enabled browser back button when full screen mode.

### DIFF
--- a/src/components/renderer/renderer.tsx
+++ b/src/components/renderer/renderer.tsx
@@ -118,6 +118,7 @@ class Editor extends React.Component<Props, State> {
     event.preventDefault();
     if (lastword === 'view') {
       this.setState({ fullscreen: false });
+      this.onClosePortal();
     }
   }
 
--- a/src/components/renderer/renderer.tsx
+++ b/src/components/renderer/renderer.tsx
@@ -116,7 +116,7 @@ class Editor extends React.Component<Props, State> {
     const urlarray = pathname.split('/');
     const lastword = urlarray[urlarray.length - 1];
     event.preventDefault();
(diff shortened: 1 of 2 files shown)


src/components/renderer/renderer.tsx
```diff
@@ -110,6 +110,17 @@ class Editor extends React.Component<Props, State> {
       .initialize(chart)
       .runAsync();
   }
+
+  public handlePopState(event) {
+    const pathname = Editor.pathname;
+    const urlarray = pathname.split('/');
+    const lastword = urlarray[urlarray.length - 1];
+    event.preventDefault();
+    if (this.state.fullscreen && lastword === 'view') {
```
domoritz: You can just check whether the string ends in "view"

takedakun:
Do you mean we should push history when we push browser back button?

domoritz:
No, I was just wondering where we push? I guess when the user opens the fullscreen modal. 

Can you update the code to check what then url ends with?

domoritz:
No, I was just wondering where we push? I guess when the user opens the fullscreen modal. 

Can you update the code to check what then url ends with?

takedakun:
I have one question. I think we can't use browser back button to get out of full screen modal when we in "/" or "/edited" url, because it only changes react state (this.state.fullscreen) and doesn't change browser history when opening full screen modal in those url. What do you think about this?

domoritz:
That's why I asked about `pushState`. Anyway, it sounds like this needs some more work and we probably want a solution that works for all routes and not just fullscreen so I'm going to close this. Please open a new PR.

Issue #128, opened 2018-07-09
Vega-Lite to Vega should open Vega spec


domoritz:
Right now we clear the spec and show an empty Vega spec.

ydlamba:
We already have a way to switch to Vega spec.

domoritz:
Yeah, but why not just open the corresponding Vega spec when you change the type?

algomaster99:
@domoritz We already have a compiled vega spec pane right? So what is the need?

domoritz:
When you switch from Vega-Lite to Vega through the dropdown, we should not clear the spec and instead open the compiled Vega spec.

PR #345, opened 2019-04-21
Vega Spec restored when switching to vg from vl in header dropdown.

ganeshpatro321:
Going to vega from vega-lite will open the compiled vega spec. #128

### DIFF
--- a/src/components/header/index.tsx
+++ b/src/components/header/index.tsx
@@ -6,7 +6,6 @@ import Renderer from './renderer';
 
 export function mapStateToProps(state: State, ownProps) {
   return {
-    compiledVegaSpec: state.compiledVegaSpec,
     editorString: state.editorString,
     lastPosition: state.lastPosition,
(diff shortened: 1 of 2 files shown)


domoritz:
I'm surprised that you added this to the components. I thought we could easily write this in the action reducers. Do you think you could try that?

src/components/header/index.tsx
```diff
@@ -6,6 +6,7 @@ import Renderer from './renderer';
 
 export function mapStateToProps(state: State, ownProps) {
   return {
+    compiledVegaSpec: state.compiledVegaSpec,
```
domoritz: Do we need this?

ganeshpatro321: No we do not, updating.

ganeshpatro321:
@domoritz Can you please squash the commits.

domoritz:
Thanks!

PR #346, opened 2019-04-22
Workbench cleanup

zhirzh:
- [x] housekeeping
- [x] replace tslint with eslint


domoritz:
Thanks! Ping me when this is ready.

zhirzh:
@domoritz adding eslint yields [500+ linting errors](https://pastebin.com/ssiwQ10K) which are mostly react specific (deprecations, best practices, code style prefs etc). After suppressing most of those rules, there are still 100+ errors.
Please have a look at them to prioritize tslint-to-eslint migration.

domoritz:
We should ignore `@typescript-eslint/explicit-function-return-type`, `@typescript-eslint/no-explicit-any`. `@typescript-eslint/no-unused-vars` could be enforced by typescript itself if we wanted to. `@typescript-eslint/explicit-member-accessibility` should be ignored for constructors, I think. I've noticed `import/no-extraneous-dependencies` producing wrong results when you import something specific and `*`. For the other ones, I trust your judgement to either disable the rule or fix the code. Does this help?

ganeshpatro321:
Hello @zhirzh 

The work for switching to eslint from tslint seems pretty good. I was working in this and faced few issues,
My only concern is whether its showing correct wanings/errors or not.

In my case there were few unexpected errors coming up:
Like I kept the no-console property on, and added some console.log() statements in the code, sometimes it was showing the error,  and sometimes it was committing changes.

Plus, when showing up the console error, it also shoed some no-unused-vars errors even if they were used.
You can check my  ```eslintrc``` file [here](https://github.com/ganeshpatro321/editor/blob/toEslint/.eslintrc.js).
And dont forget to update the scripts.

Cheers!

zhirzh:
@domoritz from what I can tell, the errors are broadly of two types:

 - actual bugs/errors
 - stylistic choices, anti-patterns and "best" practices

Since I am running prettier independent of eslint, I can say with certainity that the latter was introduced by eslint-config-airbnb. Removing just that one package reduces the number to a managable ~100 errors.

zhirzh:
Since the project uses now deprecated react features heavily (componentWillReceiveProps, findDOMNode, string refs) I suggest we rollback react to an older version where these methods are safe to use

domoritz:
Why are they not safe to use? Yes, we can change the code but I don't think going to an older version will be better.

zhirzh:
Perhaps I should've been clearer when I said "safe". What I meant was:

- React docs explicitly mark some lifecycle methods unsafe and will drop them in v17. These are:
  - UNSAFE_componentWillMount()
  - UNSAFE_componentWillUpdate()
  - UNSAFE_componentWillReceiveProps()
- React team has shown strong sentiments against [using string refs](https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs) and [overusing](https://reactjs.org/docs/refs-and-the-dom.html#dont-overuse-refs) refs in general
- Moving forward react team introduced a strict mode that [throws a warning on string refs](https://reactjs.org/docs/strict-mode.html#warning-about-legacy-string-ref-api-usage) and outright [deprecated findDomNode in strict mode](https://reactjs.org/docs/react-dom.html#finddomnode)

Maybe you're right about there being no benefit in going to an older version but I see no benefit adopting newer versions either unless the project undergoes a major overhaul.

domoritz:
I think we can update things over time to follow best practices more.

domoritz:
What's left to do before we can merge?

domoritz:
@zhirzh do you want to finish this?

zhirzh:
@domoritz yes I will. it's just that I am running busy this month. sorry for the delay.

domoritz:
No problem. It would be good if you could finish up a piece so we can merge it before we get out of sync.

domoritz:
@zhirzh @siddhant1 can we finish this up?

siddhant1:
Will try to finish this up!

siddhant1:
@domoritz @zhirzh Whats remaining in this one?
And @domoritz , react removed these lifecycle hooks because they are mostly misued and have a very specific use case. Now we have more lifecycle hooks to (kinda) replace the previous ones.

domoritz:
Set the pr to be ready, merge Master, maybe use the Vega Dev config as we have done in q or in Vega-Lite (but this could come later).

domoritz:
It would be great to use https://github.com/vega/vega-lite-dev-config here.

domoritz:
@zhirzh @ydlamba Can we finish this up so that we can merge it?

domoritz:
replaced by https://github.com/vega/editor/pull/484

PR #349, opened 2019-05-15
Fix resizing of debug pane in firefox

### LINKED ISSUE #348: Resizing of debug pane
resize

This bug persists in Firefox only.

algomaster99:
resize

Fixes #348

### DIFF
--- a/src/constants/consts.ts
+++ b/src/constants/consts.ts
@@ -21,6 +21,7 @@ export const LAYOUT = {
   DebugPaneSize: 200,
   HeaderHeight: 60,
   MinPaneSize: 30,
+  MinUpperPaneHeight: 70,
 };
 
 export type Renderer = 'svg' | 'canvas';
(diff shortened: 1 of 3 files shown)


domoritz:
Let's not make the minimum height 0 to avoid hiding the panel completely. Maybe we can limit it to 100 pixels for example.

algomaster99:
@domoritz The input panel pane also has the `min-height` 0. Shall I make it 100 too?

domoritz:
Probably. Do we have a constant for it? If not, let's make one.

src/constants/consts.ts
```diff
@@ -21,6 +21,7 @@ export const LAYOUT = {
   DebugPaneSize: 200,
   HeaderHeight: 60,
   MinPaneSize: 30,
+  MinUpperPaneHeight: 70,
```
domoritz: Let's just use `MinPaneSize`.

PR #354, opened 2019-05-24
Refactor Modal

siddhant1:
Changed the implementation of help modal to match with other portals with the state.
Earlier the portal was not closing on `esc` , but I made the implementation to match with other portals with state


domoritz:
Thank you

PR #356, opened 2019-05-26
Added Shortcut to Help Modal Portal

siddhant1:


siddhant1:
The tslint is giving some error , saying that I shadowed the event object , but in my implementation I didn't do any shadowing , I am just using the arguement.

siddhant1:
Don't merge it right now , I will add some documentation and remove some redundant checks

src/components/header/renderer.tsx
```diff
@@ -71,6 +72,30 @@ class Header extends React.Component<Props, State> {
     });
   }
 
+  public handleHelpModalToggle(event, openPortal, closePortal, isOpen) {
+    if (this._listnerAttached) {
```
algomaster99: Why do we need this condition?

src/components/header/renderer.tsx
```diff
@@ -301,26 +326,30 @@ class Header extends React.Component<Props, State> {
...
-                  <div className="modal modal-top" onClick={e => e.stopPropagation()}>
-                    <div className="modal-header">
-                      <button className="close-button" onClick={closePortal}>
-                        <X />
-                      </button>
-                    </div>
-                    <div className="modal-body">
-                      <HelpModal />
+            {({ openPortal, closePortal, isOpen, portal }) => {
+              !this._listnerAttached && this.handleHelpModalToggle(event, openPortal, closePortal, isOpen);
```
algomaster99: Can you explain what `this._listenerAttached` is ensuring?

src/components/header/renderer.tsx
```diff
@@ -71,6 +72,30 @@ class Header extends React.Component<Props, State> {
...
   }
 
+  public handleHelpModalToggle(event, openPortal, closePortal, isOpen) {
+    if (this._listnerAttached) {
+      return;
+    }
+
+    window.addEventListener('keydown', event => {
+      if (
+        (event.keyCode === 222 && event.metaKey && !event.shiftKey) || // Handle key press in Mac
```
algomaster99: Will be better if we store these key codes in `consts.ts`

siddhant1:
I need to remove some redundant checks

src/components/header/renderer.tsx
siddhant1: I am just checking if the listeners are attached , because the render props will attach listener on every render , which causes the page to crash , so.this condition just checks if the the listeners are attached , if no then call the function that attaches listeners otherwise it will just return and the attached listener will do the job .
I will document this later

algomaster99: @siddhant1 but `windows.addEventListener` will still attach the listener right?

src/components/header/renderer.tsx
siddhant1: Yes! I will do it.

src/components/header/renderer.tsx
siddhant1: The function `handleHelpModalToggle` is resposible for adding the event listener , on initial render , the events will be attached and the _listnerAttached will be set to true, now the listener wont be reattached on every render because the this._listener will resolve to true , so the expression will not to evaluated , hence no events attaching

algomaster99:
@siddhant1 Add the keycodes to `consts.ts` files.

siddhant1:
We need to refactor at every place if we want to move keyboard shortcuts in the `consts.ts` file , I feel this is okay for now and ready to be merged

siddhant1:
I looked at the travis log and there is a convention to start private variables with `_` and I have to pass a callback function to removing the event listeners .
@domoritz Can you review this

src/components/header/renderer.tsx
```diff
@@ -71,6 +72,26 @@ class Header extends React.Component<Props, State> {
...
+        } else {
+          closePortal();
+        }
+      }
+      this._listnerAttached = true;
+    });
+  }
+
+  public componentWillUnmount() {
+    window.removeEventListener('keydown', () => {});
```
algomaster99: Just write `return;` inside it so the "block empty" error resolves.

src/components/header/renderer.tsx
```diff
@@ -33,6 +33,7 @@ const formatExampleName = (name: string) => {
 class Header extends React.Component<Props, State> {
   private refGistForm: HTMLFormElement;
   private examplePortal = React.createRef<HTMLDivElement>();
+  private _listnerAttached: boolean = false;
```
algomaster99: Refer to [this](https://github.com/tc39/proposal-class-fields/blob/master/PRIVATE_SYNTAX_FAQ.md), `#` is used for declaring private variables instead of `_`. Try this if that doesn't pass the test, name what the tests are suggesting.

domoritz:
Please fix the CI

siddhant1:
This one is complete , removing the WIP , @domoritz please verify!

src/components/header/renderer.tsx
domoritz: This is still todo, right?

siddhant1:
I think there are many places where the keycodes are written like this only, what about including this in the workbench cleanup?

src/components/header/renderer.tsx
algomaster99: @siddhant1 fix more of these types of changes in workbench cleanup. It's okay if this is resolved here itself.

PR #363, opened 2019-06-22
Also export config to JSON

siddhant1:


src/components/header/export-modal/renderer.tsx
```diff
@@ -77,10 +86,15 @@ class ExportModal extends React.PureComponent<Props, State> {
     ) {
       return;
     }
-    const content = this.state.downloadVegaJSON ? this.props.vegaSpec : this.props.vegaLiteSpec;
+    let content = this.state.downloadVegaJSON ? this.props.vegaSpec : this.props.vegaLiteSpec;
     const filename = this.state.downloadVegaJSON ? `visualization.vg.json` : `visualization.vl.json`;
+    if (this.state.includeConfig) {
+      content = mergeDeep(content, { config: this.props.config });
```
domoritz: Please confirm that the spec.config has precedence over the config. Also confirm that the behavior is consistent (e.g. make a spec.config and a config that set the background color and see what happens.

siddhant1: Sure!

siddhant1: Made the change!
Findings
1. Even if the config is written , the vega automatically takes the spec.config as of higher priority.
2. The fix includes checking if the config already exists ? dont merge : merge

domoritz: This is incorrect. We specifically want to merge even if the config exists in the spec already.

siddhant1: Should we merge spec.config and config giving spec.config a priority?

domoritz: Yes (if that's consistent with Vega's behavior).

src/components/header/export-modal/renderer.tsx
```diff
@@ -77,10 +86,24 @@ class ExportModal extends React.PureComponent<Props, State> {
...
+    let config = {};
+
+    // Check if the content has the config(i.e spec.config)
+    if (content.config) {
+      // Merge the spec.config and the custom config
+      config = mergeDeep(this.props.config, content.config);
+    }
+
+    if (this.state.includeConfig) {
+      content = mergeDeep(content, { config });
```
domoritz: mergeDeep already merges into the first object no?

src/components/header/export-modal/renderer.tsx
```diff
@@ -77,10 +86,24 @@ class ExportModal extends React.PureComponent<Props, State> {
...
     }
-    const content = this.state.downloadVegaJSON ? this.props.vegaSpec : this.props.vegaLiteSpec;
+    let content = this.state.downloadVegaJSON ? this.props.vegaSpec : this.props.vegaLiteSpec;
     const filename = this.state.downloadVegaJSON ? `visualization.vg.json` : `visualization.vl.json`;
 
-    const blob = new Blob([JSON.stringify(content, null, 2)], { type: `application/json` });
+    let config = {};
+
+    // Check if the content has the config(i.e spec.config)
+    if (content.config) {
```
domoritz: We shouldn't need this since we merge below, no? Why merge twice?

siddhant1:
I think we are firstly merging both the configs and then merging config with the spec

domoritz:
I don't see why that is necessary.

domoritz:
I think `spec = deepMerge({config}, spec)` is sufficient, no?

siddhant1:
@domoritz I think we can finalize this implementation

src/components/header/export-modal/renderer.tsx
```diff
@@ -80,7 +89,15 @@ class ExportModal extends React.PureComponent<Props, State> {
     const content = this.state.downloadVegaJSON ? this.props.vegaSpec : this.props.vegaLiteSpec;
     const filename = this.state.downloadVegaJSON ? `visualization.vg.json` : `visualization.vl.json`;
 
-    const blob = new Blob([JSON.stringify(content, null, 2)], { type: `application/json` });
+    if (this.state.includeConfig) {
+      const Newcontent = mergeDeep({ config: this.props.config }, content as any);
```
domoritz: Variables shouldn't start with capital letters. 

Why didn't the code I suggested not work?

siddhant1: 1. Sure I will change the name of the variable.
2. The second argument takes a priority , so If is pass custom config as first arguement and content (spec.config) as second arguement , the result is perfect , but the config (key) appears on top of the json.
That is why I have deleted the key and added it again , so is appears at the bottom

domoritz: I see. Can you add a comment to explain that?

siddhant1: Sure!

domoritz: I don't see the comment. Can you only respond to my comments once you finished the task? It makes it easier for me to track what has happened and what still needs to be done. Okay?

siddhant1: Okay! pushed changes. Should I make a branch for you to test?

PR #364, opened 2019-06-24
Add authentication and loading gists

algomaster99:
Here is how the authentication button will look:
auth

The PR will be ready for merge after:
- [x] The profile image is sent by back-end
- [ ] I have currently removed the docs tab. Needs to be adjusted somewhere.


algomaster99:
@domoritz Please fetch this and test. The gif didn't upload correctly :confused:

domoritz:
Instead of using the three dots, can we just make the profile image the button? Can you also make the popup a bit larger so it looks nicer/ more consistent?

src/components/header/renderer.tsx
```diff
@@ -19,10 +20,13 @@ type Props = ReturnType<typeof mapStateToProps> &
   ReturnType<typeof mapDispatchToProps> & { history: any; showExample: boolean };
 
 interface State {
+  open: boolean;
   showVega: boolean;
   scrollPosition: number;
 }
 
+const url = 'https://vega.now.sh/';
```
domoritz: Use all caps and use a more descriptive name.

src/components/header/renderer.tsx
```diff
@@ -171,6 +213,25 @@ class Header extends React.PureComponent<Props, State> {
       </div>
     );
 
+    const auth = this.props.isAuthenticated ? (
```
domoritz: Can we make the profile button a new component?

src/components/header/renderer.tsx
```diff
@@ -434,11 +535,9 @@ class Header extends React.PureComponent<Props, State> {
             ]}
           </PortalWithState>
 
-          <span>{docsLink}</span>
+          {/* <span>{docsLink}</span> */}
```
domoritz: Can we leave this in for now and reduce the spacing between buttons?

algomaster99:
@domoritz 
### Logged In
Screenshot from 2019-07-12 21-32-18

### Logged Out
Screenshot from 2019-07-12 22-24-47

algomaster99:
@domoritz Do you have any idea why travis is failing?

domoritz:
Something is broken. I'm looking.

domoritz:
The "log out" menu doesn't look like the drop down for the other menu items. The log in with black background looks a bit out of place. Have a look at https://observablehq.com/ for inspiration.

src/components/header/renderer.tsx
```diff
@@ -165,6 +203,35 @@ class Header extends React.PureComponent<Props, State> {
...
+                    <input className="sign-out" type="submit" value="Logout" onClick={e => e.stopPropagation()} />
+                  </div>
+                </div>
+              )}
+            </div>
+          </form>
+        ) : (
+          <form action={`${BACKEND_URL}auth/github`} method="get">
+            <button className="sign-in" type="submit">
+              <span className="sign-in-text">Login with</span>
```
domoritz:
```suggestion
              <span className="sign-in-text">Sign in with</span>
```

src/components/header/renderer.tsx
```diff
@@ -165,6 +203,35 @@ class Header extends React.PureComponent<Props, State> {
       </div>
     );
 
+    const authButton = (
+      <div className="auth-button-container">
+        {this.props.isAuthenticated ? (
+          <form action={`${BACKEND_URL}auth/github/logout`} method="get">
+            <div className="profile-container">
```
domoritz: That's a lot of classes. Do we need them all?

src/constants/consts.ts
```diff
@@ -53,3 +53,7 @@ export const KEYCODES = {
   SINGLE_QUOTE: 222,
   SLASH: 191,
 };
+
+export const BACKEND_URL = 'https://vega.now.sh/';
+
+export const COOKIE_NAME = 'vegasessid';
```
domoritz:
```suggestion
export const COOKIE_NAME = 'vega_session';
```

src/constants/default-state.ts
```diff
@@ -5,6 +5,7 @@ import { LocalLogger } from '../utils/logger';
 import { LAYOUT, Mode, NAVBAR, Renderer, SIDEPANE, VEGA_LITE_START_SPEC, View } from './consts';
 
 export interface State {
+  isAuthenticated: boolean;
```
domoritz: Can we even put this in the state that is being serialized? I feel we might need to check this when the user is opening the editor. Or am I missing something?

src/utils/getCookie.ts
```diff
@@ -0,0 +1,17 @@
+const getCookie = cname => {
```
domoritz: Let's make this a real function.

src/components/header/index.css
```diff
@@ -297,8 +297,123 @@
...
+.welcome {
+  font-size: 0.7em;
+  color: #6e6e6e;
+  margin: 1em 0 0.5em 0.6em;
+}
+
+.whoami {
+  font-size: 0.9em;
+  padding: 0.5em 0 1em 0.5em;
+  /* padding-bottom: 0.5em; */
```
domoritz: delete

domoritz:
- [ ] Add description to the readme about how to test this feature.

domoritz:
- [ ] Make sure that we don't have unsaved changes before navigating away. Can we somehow sign in in a new tab rather than reloading the page?

src/components/header/renderer.tsx
```diff
@@ -165,6 +203,35 @@ class Header extends React.PureComponent<Props, State> {
...
+              <img className="profile-img" src={this.props.profilePicUrl} />
+              <span className="arrow-down"></span>
+              {this.state.open && (
+                <div className="profile-menu">
+                  <div className="welcome">Welcome!</div>
+                  <div className="whoami">{this.props.name}</div>
+                  <div className="profile-options">
+                    <input className="sign-out" type="submit" value="Logout" onClick={e => e.stopPropagation()} />
```
domoritz:
```suggestion
                    <input className="sign-out" type="submit" value="Sign out" onClick={e => e.stopPropagation()} />
```

src/components/header/renderer.tsx
```diff
@@ -165,6 +203,35 @@ class Header extends React.PureComponent<Props, State> {
...
+    const authButton = (
+      <div className="auth-button-container">
+        {this.props.isAuthenticated ? (
+          <form action={`${BACKEND_URL}auth/github/logout`} method="get">
+            <div className="profile-container">
+              <img className="profile-img" src={this.props.profilePicUrl} />
+              <span className="arrow-down"></span>
+              {this.state.open && (
+                <div className="profile-menu">
+                  <div className="welcome">Welcome!</div>
```
domoritz:
```suggestion
                  <div className="welcome">Logged in as</div>
```

domoritz:
Screen Shot 2019-08-11 at 1 55 16 PM

* Make the rule less prominent
* Add more padding around the popup
* Move the popup up
* Make it somehow clear (visually) that you cannot click on the user name

domoritz:
Clicking on the user name should open the model that shows the user's visualizations (the gist dialog for now).

src/constants/default-state.ts
algomaster99: @domoritz Yes, we check this. Doesn't it automatically store it in localStorage?

domoritz: It does but my question is whether that is right. If we serialize this and then the user opens the editor after 10 years, we will tell them that they are logged in even if the login has timed out.

src/components/header/renderer.tsx
algomaster99: @domoritz yeah, the component is a bit complicated

src/constants/default-state.ts
algomaster99: @domoritz Whenever the editor is opened, a fetch request is made to the back-end to check if the user is logged in and accordingly the state is set.

src/utils/getCookie.ts
algomaster99: @domoritz Can you elaborate? I don't get you. :/

domoritz: If you can, use `function` instead of making this a `const gteCookie`.

src/constants/default-state.ts
domoritz: Okay, then let's not persists `isAuthenticated`. Can you exclude it?

algomaster99:
@domoritz We can sign-in in a new tab but what are the unsaved changes? Editor most of the times save things automatically.

domoritz:
For example, the interactive state of a visualization is not saved.

src/components/header/renderer.tsx
algomaster99: @domoritz I removed a class called `profile-options`.

domoritz:
- [ ] Don't make the logout text blue

domoritz:
Missing fix for

Screen Shot 2019-08-18 at 10 16 30 AM

PR #366, opened 2019-06-29
Coalesce share modal into export modal

algomaster99:
### Demo
pop


domoritz:
Thank you for merging the two. Do you think we should make sharing by url the first entry instead of the last?

algomaster99:
@domoritz I think it's fine. Their equal size assert equal importance.

domoritz:
I think people still gravitate to look at the first option first. I'd like sharing to be the first option.

algomaster99:
@domoritz 
Screenshot from 2019-06-30 21-58-32

domoritz:
Thank you!

PR #376, opened 2019-07-06
Fixed empty config issue

### LINKED ISSUE #374: Cannot remove theme
When I delete the config entirely, the chart doesn't automatically update to be without the config. The only way to reset the config is to type `{}`.

siddhant1:
Fixes #374


domoritz:
Please use comments to say that this fixes #374 so that github closes the issue automatically.

src/components/config-editor/index.tsx
```diff
@@ -9,7 +9,11 @@ import './config-editor.css';
 
 class ConfigEditor extends React.PureComponent<any, any> {
   public handleEditorChange = spec => {
-    this.props.setConfig(JSON.parse(spec));
+    if (spec === '') {
+      this.props.setConfig({});
```
domoritz: SO am I right to assume that this also sets the editor content? Then don't we have a loop in the editor -> config -> editor?

siddhant1: Maybe we will have a loop! but that also will happen only one time because on 1st loop it will set the config to `{}` and then the config won't change.

I am checking if that's also happening

domoritz: We have a loop. Can you fix it? 

editor

siddhant1: Checked! The onChange handler only works when the user change the config! It does not work when the config is changed programatically

siddhant1:
@domoritz I feel this is complete now

domoritz:
I two bugs. 

- [ ] In manual mode, changing the spec should not update the chart
- [ ] When I set a config, delete everything, the dropdown switches to custom, but when I then open the Vega tab and then go back to config, the previous theme is selected

And one feature that would be nice.

- [ ] Support autocomplete. I don't know what a good way to do this is, though. If the user is on the latest Vega(-Lite) version that we use in the editor, then we can use the schema we have from npm. I think this takes a bit more time to implement so please file an issue for this if you don't implement it.

src/components/config-editor/index.tsx
```diff
@@ -10,6 +10,7 @@ import './config-editor.css';
 class ConfigEditor extends React.PureComponent<any, any> {
   public handleEditorChange = spec => {
     this.props.setConfigEditorString(spec);
+    this.props.setThemeName('custom');
```
domoritz: please rename

siddhant1: What should be renamed?

domoritz: Hmm, I guess we can call it theme here. Never mind.

siddhant1:
@domoritz can you check!

siddhant1:
Can you check if this is the desired behavior! I have opened a branch in vega/editor

domoritz:
Please merge master.

src/components/header/renderer.tsx
```diff
@@ -165,7 +165,18 @@ class Header extends React.PureComponent<Props, State> {
     );
 
     const runButton = (
-      <div className="header-button" onClick={() => this.props.parseSpec(true)}>
+      <div
+        className="header-button"
+        id="run-button"
+        onClick={() => {
+          if (this.props.configEditorString === '') {
```
domoritz: Please move this logic into the `parseSpec` action.

src/components/header/renderer.tsx
```diff
@@ -188,6 +199,7 @@ class Header extends React.PureComponent<Props, State> {
                     <div
                       key={j}
                       onClick={() => {
+                        document.getElementById('run-button').click();
```
domoritz: This is not acceptable. We shouldn't be clicking buttons just to get things working. Look at the actions for loading a spec and the renderer.

src/components/config-editor/config-editor-header.tsx
```diff
@@ -19,8 +20,12 @@ class ConfigEditorHeader extends React.PureComponent<any, any> {
           id="config-select"
           onChange={e => {
             e.stopPropagation();
-            this.props.setConfig(themes[e.target.value]);
+            this.props.setConfigEditorString(stringify(themes[e.target.value]));
```
domoritz: Why not just introduce a `setTheme`, which then also sets the config?

src/components/app.tsx
```diff
@@ -148,6 +149,7 @@ class App extends React.PureComponent<Props & { match: any; location: any; showE
 
 function mapStateToProps(state: State) {
   return {
+    configEditorString: state.configEditorString,
```
domoritz: This mapStateToProps shouldn't even be here.

src/components/config-editor/config-editor-header.tsx
```diff
@@ -49,6 +52,7 @@ export function mapDispatchToProps(dispatch: Dispatch<EditorActions.Action>) {
 
 function mapStateToProps(state) {
   return {
+    manualParse: state.manualParse,
```
domoritz: This mapStateToProps shouldn't even be here.

siddhant1: Where should I keep it then?

siddhant1: we can remove the manual parse but I think we need the themeName in this case

domoritz: Check my commits from today.

PR #377, opened 2019-07-06
Add svg in config-header

siddhant1:


src/components/input-panel/compiled-spec-header/index.tsx
```diff
@@ -62,7 +62,9 @@ class CompiledSpecDisplayHeader extends React.PureComponent<Props> {
             </li>
           </ul>
 
-          <ChevronDown />
+          <div className="down-button-svg">
```
domoritz: Why is this necessary?

siddhant1: I tried `.editor-header svg` , its not working for me

domoritz: I'm not really satisfied by that answer.

siddhant1: svg is an inline-element , we can place it inside div and position the div , otherwise if we want to do it with only SVG tag, we have to position it using pixels

domoritz: Really? Can't we just make it a block element for example?

siddhant1: Tried! It's actually not working , maybe when we switch we remove some divs!
Np, I am trying to place it without the div , but I think this works good as well!

PR #378, opened 2019-07-07
Isolate and revise all key codes used

algomaster99:


algomaster99:
@domoritz I am not authorized to merge.

PR #380, opened 2019-07-08
Removed extra margin from help-icon

siddhant1:


src/components/header/renderer.tsx
```diff
@@ -134,7 +134,7 @@ class Header extends React.PureComponent<Props, State> {
 
     const HelpButton = (
       <div className="header-button help" onClick={() => this.setState(current => ({ ...current }))}>
```
domoritz: We already have the class `help` here. No need for another class.

siddhant1: fixed!

siddhant1:
Should I merge?

domoritz:
Yes

PR #386, opened 2019-07-13
Correct type of error

domoritz:
We previously assigned a string to the `error` property, which is supposed to be an `Error`.


domoritz:
Can you look at the Travis logs?

siddhant1:
Build is failing , even on master

PR #387, opened 2019-07-14
Add support for vega-lite config

### LINKED ISSUE #375: Config does not work correctly for Vega-Lite
Try adding `{"view": {"fill": "red"}}` to a config. It doesn't make background red.

siddhant1:
Fixes #375


src/components/renderer/renderer.tsx
```diff
@@ -57,8 +58,10 @@ class Editor extends React.PureComponent<Props, State> {
 
   // Initialize the view instance
   public initView() {
-    const runtime = vega.parse(this.props.vegaSpec, this.props.config);
-
+    let runtime = vega.parse(this.props.vegaSpec, this.props.config);
```
domoritz: So we always parse here first? Make this dependent on the mode.

src/components/renderer/renderer.tsx
```diff
@@ -57,8 +58,10 @@ class Editor extends React.PureComponent<Props, State> {
...
+    if (this.props.mode === Mode.VegaLite) {
+      runtime = vega.parse(this.props.vegaSpec);
```
domoritz: Add a comment where the config comes in.

src/reducers/index.ts
```diff
@@ -105,19 +105,31 @@ function parseVega(
...
-  action: SetVegaLiteExample | UpdateVegaLiteSpec | SetGistVegaLiteSpec,
-  extend: Partial<State> = {}
-) {
+function parseVegaLite(state: State, action: any, extend: Partial<State> = {}) {
   const currLogger = new LocalLogger();
 
   try {
+    // if we are setting the spec we don't pass the config
+    // add the previous config from the state
+    if (!action.config) {
```
domoritz: This is causing side effects. Instead, rewrite the `options` code below to use `action.config || state.config`.

src/reducers/index.ts
```diff
@@ -105,19 +105,31 @@ function parseVega(
...
+      action.config = state.config;
+    }
+
+    // if we are setting the config we don't pass the spec
+    // add the previous spec from the state
+    if (!action.spec) {
```
domoritz: same as above

siddhant1: image
We have to do this because of this error

siddhant1: JSON.parse(undefined||"{}") throws an error

siddhant1: Works for object though!

src/reducers/index.ts
```diff
@@ -105,19 +105,22 @@ function parseVega(
...
-) {
+function parseVegaLite(state: State, action: any, extend: Partial<State> = {}) {
   const currLogger = new LocalLogger();
 
   try {
-    const spec = JSON.parse(action.spec);
+    if (!action.spec) {
+      action.spec = state.editorString;
+    }
+    const spec = JSON.parse(action.spec || state.editorString);
```
domoritz: Wait, you already assigned actions.spec. This is redundant.

src/reducers/index.ts
```diff
@@ -105,19 +105,22 @@ function parseVega(
...
-  state: State,
-  action: SetVegaLiteExample | UpdateVegaLiteSpec | SetGistVegaLiteSpec,
-  extend: Partial<State> = {}
-) {
+function parseVegaLite(state: State, action: any, extend: Partial<State> = {}) {
   const currLogger = new LocalLogger();
 
   try {
+    if (!action.spec) {
+      action.spec = state.editorString;
```
domoritz: This also causes a side effect that we don't want.

siddhant1: What is the side effect?

domoritz: The code modifies an input. A function should generally avoid doing that.

src/reducers/index.ts
```diff
@@ -105,19 +105,22 @@ function parseVega(
   };
 }
 
-function parseVegaLite(
-  state: State,
-  action: SetVegaLiteExample | UpdateVegaLiteSpec | SetGistVegaLiteSpec,
-  extend: Partial<State> = {}
-) {
+function parseVegaLite(state: State, action: any, extend: Partial<State> = {}) {
```
domoritz: Don't use any here.

src/reducers/index.ts
```diff
@@ -105,19 +105,22 @@ function parseVega(
...
   const currLogger = new LocalLogger();
 
   try {
+    if (!action.spec) {
+      action.spec = state.editorString;
+    }
     const spec = JSON.parse(action.spec);
```
domoritz:
```suggestion
    const spec = JSON.parse(action.spec || state.editorString;);
```

domoritz:
Please fix the tests on travis.

domoritz:
Tests are still failing

package.json
```diff
@@ -19,7 +19,7 @@
     "tslint": "^5.18.0",
     "tslint-config-prettier": "^1.18.0",
     "tslint-eslint-rules": "^5.4.0",
-    "typescript": "^3.5.3",
+    "typescript": "~3.4.0",
```
algomaster99: @siddhant1 Why are we downgrading the version?

src/components/header/share-modal/renderer.tsx
```diff
@@ -25,8 +27,15 @@ class ShareModal extends React.PureComponent<Props, State> {
...
+    let spec;
+    if (this.props.mode === Mode.VegaLite) {
+      spec = this.props.vegaLiteSpec;
+    } else {
+      spec = this.props.vegaSpec;
+    }
+    spec.config = this.props.config;
     const serializedSpec =
-      LZString.compressToEncodedURIComponent(this.props.editorString) + (this.state.fullScreen ? '/view' : '');
+      LZString.compressToEncodedURIComponent(stringify(spec)) + (this.state.fullScreen ? '/view' : '');
```
algomaster99: @siddhant1 I suggest using template literals.

package.json
domoritz: I did that because Vega-Lite is not yet TS 3.5 ready.

src/components/header/share-modal/renderer.tsx
domoritz: This is actually not what we want. We do want to save the spec like it was formatted. 

If you really feel like changing this, then we should store the spec compressed without whitespace.

src/components/renderer/renderer.tsx
```diff
@@ -57,8 +58,13 @@ class Editor extends React.PureComponent<Props, State> {
 
   // Initialize the view instance
   public initView() {
-    const runtime = vega.parse(this.props.vegaSpec, this.props.config);
-
+    let runtime = null;
```
domoritz: Add a type here.

src/components/renderer/renderer.tsx
```diff
@@ -58,7 +58,7 @@ class Editor extends React.PureComponent<Props, State> {
 
   // Initialize the view instance
   public initView() {
-    let runtime = null;
+    let runtime: Runtime;
```
domoritz: use `vega.Runtime`

src/reducers/index.ts
```diff
@@ -107,25 +111,43 @@ function parseVega(
...
+        break;
+      case SET_VEGA_LITE_EXAMPLE:
+      case SET_GIST_VEGA_LITE_SPEC:
+      case UPDATE_VEGA_LITE_SPEC:
+        spec = action.spec;
+        config = state.config as Config;
+    }
 
-    validateVegaLite(spec, currLogger);
+    const vegaLiteSpec: TopLevelSpec = JSON.parse(spec);
```
siddhant1: Should I use vl.TopLevelSpec here as well?

domoritz:
- [ ] Switching to VG via the "Edit Vega Spec" button doesn't delete the config.

src/components/header/share-modal/renderer.tsx
```diff
@@ -1,9 +1,11 @@
+import stringify from 'json-stringify-pretty-compact';
 import LZString from 'lz-string';
 import * as React from 'react';
 import Clipboard from 'react-clipboard.js';
 import { Copy, Link } from 'react-feather';
 import { withRouter } from 'react-router-dom';
 import { mapStateToProps } from '.';
+import { Mode } from '../../../constants';
```
domoritz: Why?

PR #406, opened 2019-07-24
Open vega-tooltip link in a new tab

algomaster99:
I think it would be better to open the [vega-tooltip](https://github.com/vega/vega-tooltip) link in a new tab.


domoritz:
Looks good. We should do the same for the links to docs and to hover.

domoritz:
@siddhant1 can you check all links and see whether they should be opened in a new tab?

siddhant1:
Sure! I will check the remaining

PR #415, opened 2019-07-27
Remove UI logic from CDU

### LINKED ISSUE #156: Refactor sharing modal

siddhant1:
Closes #156

### DIFF
--- a/src/components/header/share-modal/renderer.tsx
+++ b/src/components/header/share-modal/renderer.tsx
@@ -39,9 +39,18 @@ class ShareModal extends React.PureComponent<Props, State> {
   }
 
   public onCopy() {
-    this.setState({
-      copied: true,
-    });
+    if (!this.state.copied) {
+      this.setState(
+        {
+          copied: true,
+        },
+        () => {
+          setTimeout(() => {
+            this.setState({ copied: false });
+          }, 2500);
+        }
+      );
+    }
   }
 
   public handleCheck(event) {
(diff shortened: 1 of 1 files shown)


src/components/header/share-modal/renderer.tsx
```diff
@@ -39,9 +39,18 @@ class ShareModal extends React.PureComponent<Props, State> {
...
   public onCopy() {
-    this.setState({
-      copied: true,
-    });
+    if (!this.state.copied) {
+      this.setState(
+        {
+          copied: true,
+        },
+        () => {
```
domoritz: this is called when copied is false

src/components/header/share-modal/renderer.tsx
```diff
@@ -54,14 +63,6 @@ class ShareModal extends React.PureComponent<Props, State> {
     this.exportURL();
   }
 
-  public componentDidUpdate(prevProps, prevState) {
-    if (this.state.copied) {
-      setTimeout(() => {
```
domoritz: this is called when copied is true

are you sure this is right?

src/components/header/share-modal/renderer.tsx
siddhant1: Yes , this is right , the callback is run after setting the state so this is the logic
1. Set copied to true.
2. Run a timer to make it false.(we know it's true because we have set the state).
3. If copied is false , then only make it true.

PR #417, opened 2019-07-29
Refactored fullscreen navigation

### LINKED ISSUE #292: Back button not redirecting when in full-screen mode

siddhant1:
Closes #292

### DIFF
--- a/src/components/renderer/renderer.tsx
+++ b/src/components/renderer/renderer.tsx
@@ -14,7 +14,8 @@ import './index.css';
 // Add additional projections
 addProjections(vega.projection);
 
-type Props = ReturnType<typeof mapStateToProps> & ReturnType<typeof mapDispatchToProps> & { history: any };
+type Props = ReturnType<typeof mapStateToProps> &
+  ReturnType<typeof mapDispatchToProps> & { history: any } & { location: { pathname: string } };
 
 const defaultState = { fullscreen: false };
 
@@ -23,6 +24,7 @@ type State = Readonly<typeof defaultState>;
 class Editor extends React.PureComponent<Props, State> {
   public static pathname: string;
   public readonly state: State = defaultState;
+  public unlisten: any;
 
   constructor(props) {
     super(props);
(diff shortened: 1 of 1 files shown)


src/components/renderer/renderer.tsx
```diff
@@ -139,7 +140,24 @@ class Editor extends React.PureComponent<Props, State> {
...
     }
+
+    window.addEventListener(
+      'popstate',
+      event => {
+        const pathname = window.location.href;
+        const urlarray = pathname.split('/');
+        const lastword = urlarray[urlarray.length - 1];
+        event.preventDefault();
+        if (lastword === 'view') {
```
domoritz: pathname.endsWith('view')

src/components/renderer/renderer.tsx
```diff
@@ -139,7 +140,24 @@ class Editor extends React.PureComponent<Props, State> {
     if (params[params.length - 1] === 'view') {
       this.setState({ fullscreen: true });
     }
+
+    window.addEventListener(
+      'popstate',
```
domoritz: What is this event and why can't we move this code where this event is triggered?

siddhant1: This event is triggered on back button press and that is handled internally by browser

domoritz: But doesn't react route listen to this already? Can't we use that?

src/components/renderer/renderer.tsx
```diff
@@ -23,6 +24,7 @@ type State = Readonly<typeof defaultState>;
 class Editor extends React.PureComponent<Props, State> {
   public static pathname: string;
   public readonly state: State = defaultState;
+  public unlisten: any;
```
domoritz: Please add a type

siddhant1: Do we have a type for it? Should I just use `() => void`

siddhant1: Also , we do not use @types/react-router-dom in the app

domoritz: Please include it then.

src/components/renderer/renderer.tsx
domoritz: I added it. Please update the types.

src/components/renderer/renderer.tsx
domoritz: I added it to master.

PR #420, opened 2019-07-30
Changed config position

### LINKED ISSUE #372: Revise the layout for vega /config
Putting config in the same location as compiled vega is a bit confusing.

Basically, we have two inputs (the vl spec and config) and vega is the output.   Putting one of the input in the place of the output mixes two things together.  

In the future, once we have https://github.com/vega/vega-lite-to-api for getting vega-lite-api and altair outputs, then it will be quite confusing.  (It would be natural to group vega output with vega-lite-api and altair, but not with the config.)

siddhant1:
Closes #372

### DIFF
--- a/src/components/input-panel/spec-editor-header/index.css
+++ b/src/components/input-panel/spec-editor-header/index.css
@@ -1,3 +1,3 @@
 .spec-editor-header {
-  cursor: unset !important;
+  text-transform: uppercase;
 }
--- a/src/components/input-panel/spec-editor-header/index.css
+++ b/src/components/input-panel/spec-editor-header/index.css
@@ -1,0 +1,3 @@
+.spec-editor-header {
+  cursor: unset !important;
+}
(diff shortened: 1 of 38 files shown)


domoritz:
We better use the upload and download icons now, I think.

siddhant1:
Yup! I will resolve the conflicts.

domoritz:
- [ ] Why can I click on the header with VL and config? That should be removed.
- [ ] Merge and extract config are gone

src/components/input-panel/spec-editor-header/index.css
```diff
@@ -0,0 +1,3 @@
+.spec-editor-header {
+  cursor: unset !important;
```
domoritz: Don't use important. Make sure the default is not a pointer.

siddhant1: We need pane-header styles which also has higher priority cursor:pointer applied to it, so we need to mark this important otherwise these styles won't kick in

PR #427, opened 2019-08-02
Fix positioning of report button

algomaster99:


siddhant1:
image
Good most of the time but not sure abut larger res, can we not do it with position:absolute and parent being relative

algomaster99:
@siddhant1 I will try doing it with `position: absolute` and `position: relative`.

domoritz:
Let's use absolute positioning. It's more reliable.

algomaster99:
@siddhant1 Your suggestion to remote `position: relative` from `.modal-body` worked. Thanks!

PR #437, opened 2019-08-06
Add search for fields in data viewer and signals in signal viewer

siddhant1:


domoritz:
Let's add the same for the signal viewer.

src/components/table/renderer.tsx
```diff
@@ -14,7 +15,7 @@ const MAX_LENGTH = 150;
 export default class Table extends React.PureComponent<Props> {
   public render() {
     const headerNodes = this.props.header.map(h => (
-      <th title={h} key={h}>
+      <th onClick={() => this.props.onClickHandler && this.props.onClickHandler(h)} title={h} key={h}>
```
domoritz: When would the handler not be defined?

src/components/viz-pane/renderer.tsx
```diff
@@ -57,6 +103,24 @@ export default class VizPane extends React.PureComponent<Props> {
       this.props.showLogs(true);
     }
   }
+
+  public getComponent() {
```
domoritz: Which component?

src/components/table/renderer.tsx
siddhant1: When we want to reuse table without the debugger.

src/components/viz-pane/renderer.tsx
siddhant1: I just moved the position of the function , it is used to get the data-viewer,signal-viewer and the error based on which is selected

src/components/table/renderer.tsx
domoritz: Okay

src/components/viz-pane/renderer.tsx
domoritz: I think it needs a qualifier in the name,

siddhant1:
Github diff caculations is definitely struggling with this one , I moved positions of the getComponent method just because I wanted to keep it above the render method

domoritz:
still todo: https://github.com/vega/editor/pull/437#discussion_r311030998

src/components/viz-pane/renderer.tsx
siddhant1: What should be the name?

domoritz: You decide

siddhant1: Something like getVizPane Component will do

domoritz: It's not a visualization pane, right? It's a `getContextViewer` or something. Describe in the comment that it returns the debug pane, data or signal viewer.

siddhant1: Should I name it getContextViewer and add a comment explaining what it returns?

domoritz: 👍

domoritz:
* [ ] Search in compiled Vega spec
* [ ] Try to highlight only the matching string (works in https://vega.github.io/vega-editor/)
* [ ] Try to support searching in expressions (works in https://vega.github.io/vega-editor/)
* [ ] Fix bug with missing highlight when I cycle through matches (or remove cycling)

Screen Shot 2019-08-09 at 10 37 21 AM

domoritz:
- [ ] use correct word separators for json and add link

src/components/viz-pane/renderer.tsx
```diff
@@ -1,46 +1,72 @@
...
-          return <DataViewer />;
-        case NAVBAR.SignalViewer:
-          return <SignalViewer />;
-        default:
-          return null;
+
+  public onClickHandler(header: string) {
+    const editor = this.props.editorRef;
+    const model = editor.getModel();
+    // wordSeparator:  /(-?\d*\.\d\w*)|([^\[\{\]\}\:\"\,\s]+)/g
```
domoritz: clean this up

src/components/viz-pane/renderer.tsx
```diff
@@ -1,46 +1,72 @@
...
+    const range = model.findMatches(header, true, true, true, '/(-?d*.dw*)|([^[{]}:",s]+)/g', true)[0]
+      ? model.findMatches(header, true, true, true, '/(-?d*.dw*)|([^[{]}:",s]+)/g', true)[0].range
```
domoritz: Why find matches twice?

src/components/viz-pane/renderer.tsx
```diff
@@ -57,6 +83,25 @@ export default class VizPane extends React.PureComponent<Props> {
       this.props.showLogs(true);
     }
   }
+
+  public getContextViewer() {
+    // The function returns the Component to be rendered in the Context Viewer
```
domoritz: Make this a proper function documentation with `/** ... */`

src/components/viz-pane/renderer.tsx
```diff
@@ -1,46 +1,58 @@
...
-    } else {
-      return null;
+
+  public onClickHandler(header: string) {
+    const editor = this.props.editorRef;
+    const model = editor.getModel();
+    // turn on whole word search for the editor
+    !editor.getContribution('editor.contrib.findController')._state.wholeWord &&
+      editor.getContribution('editor.contrib.findController').toggleWholeWords();
+    // wordSeparator:  `~!@#$%^&*()-=+[{]}\\|;:\'",.<>/?
```
domoritz: Please import https://github.com/microsoft/monaco-json/blob/master/src/jsonMode.ts#L43

src/components/viz-pane/renderer.tsx
```diff
@@ -57,6 +73,27 @@ export default class VizPane extends React.PureComponent<Props> {
       this.props.showLogs(true);
     }
   }
+
+  /**
+   *  returns the Component to be rendered in the Context Viewer
```
domoritz:
```suggestion
   *  Get the Component to be rendered in the Context Viewer.
```

src/components/input-panel/compiled-spec-display/index.tsx
```diff
@@ -38,12 +43,13 @@ class CompiledSpecDisplay extends React.PureComponent<Props> {
...
           language="json"
           value={stringify(this.props.value)}
-          editorDidMount={e => {
-            e.onDidFocusEditorText(() => {
-              this.props.setEditorReference(this.refs.compiledEditor);
+          editorDidMount={editor => {
+            editor.onDidFocusEditorText(() => {
+              this.props.compiledEditorRef && this.props.compiledEditorRef.deltaDecorations(this.props.decorations, []);
+              this.props.editorRef.deltaDecorations(this.props.decorations, []);
+              localStorage.setItem('find_editor', 'compiled');
```
domoritz: Can we put that in the state?

siddhant1: Yeah , I will put that in the state!

domoritz: Thanks!

src/actions/editor.ts
```diff
@@ -318,6 +324,15 @@ export function setEditorReference(editorRef: any) {
 
 export type SetEditorReference = ReturnType<typeof setEditorReference>;
 
+export function setCompiledEditorRef(editorRef: any) {
```
domoritz: Can we not use any?

src/actions/editor.ts
```diff
@@ -318,6 +324,15 @@ export function setEditorReference(editorRef: any) {
...
+  return {
+    editorRef: (editorRef as any).editor,
```
domoritz: remove as any

src/constants/consts.ts
```diff
@@ -57,6 +62,11 @@ export const KEYCODES = {
   SLASH: 191,
 };
 
+/**
+ * Ref : https://github.com/microsoft/monaco-editor/blob/02fb6bfd43b5ddcff36c1a1303e4aac623deff2e/monaco.d.ts#L2602
```
domoritz:
```suggestion
 * Taken from: https://github.com/microsoft/monaco-editor/blob/02fb6bfd43b5ddcff36c1a1303e4aac623deff2e/monaco.d.ts#L2602
```

siddhant1:
If we don't wrap for large tables , we will get scroll , so should we wrap or not?

package.json
```diff
@@ -15,26 +15,29 @@
     "postcss-loader": "^3.0.0",
     "prettier": "^1.18.2",
     "style-loader": "^1.0.0",
-    "tslint-eslint-rules": "^5.4.0",
     "tslint": "^5.19.0",
+    "tslint-eslint-rules": "^5.4.0",
```
domoritz: Don't add this back

package.json
```diff
@@ -15,26 +15,29 @@
...
+    "webpack-dev-server": "^3.8.0"
   },
   "dependencies": {
+    "@types/react": "^16.9.2",
     "@types/react-redux": "^7.1.2",
     "@types/react-router-dom": "^4.3.5",
-    "@types/react": "^16.9.2",
     "ajv": "^6.10.2",
     "d3-geo-projection": "^2.7.0",
+    "eslint": "^6.2.2",
```
domoritz: remove

package.json
```diff
@@ -15,26 +15,29 @@
...
+    "global": "^4.4.0",
```
domoritz: what's this for?

src/actions/editor.ts
domoritz: Still todo

tslint.json
```diff
@@ -0,0 +1,18 @@
+{
```
domoritz: don't add this back

PR #438, opened 2019-08-06
Refactored run button styles

### LINKED ISSUE #436: Run button has too little space on small screens
There needs to be more padding on the right. 

Screen Shot 2019-08-06 at 10 20 16 AM

siddhant1:
Closes #436

### DIFF
--- a/src/components/header/index.css
+++ b/src/components/header/index.css
@@ -208,6 +208,7 @@
 
 .auto-run__menu {
   margin-top: 10px !important;
+  width: 80px !important;
 }
 
 .auto-run__option {
--- a/src/components/header/index.css
+++ b/src/components/header/index.css
@@ -274,7 +274,7 @@
 }
 
 .split-button {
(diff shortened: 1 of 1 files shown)


domoritz:
I think 

```css
.split-button {
    width: 80px;
}
```

is sufficient, no?

siddhant1:
Doesn't work for me! it messes up the styles

siddhant1:
yup! it works

domoritz:
I'd like to encourage you to try to fix things by deleting code/modifying code rather than adding more ;-)

siddhant1:
Prev:
image
After:
image

domoritz:
Why do we need the important? I didn't need it when I modified things live in the browser.

siddhant1:
It was already there in media queries  , so I thought it had a purpose! Should I remove both of them?

domoritz:
I guess it's needed.

PR #459, opened 2019-08-16
Add timeline of signals

domoritz:


src/components/signal-viewer/renderer.tsx
```diff
@@ -1,37 +1,244 @@
...
+import { mapDispatchToProps, mapStateToProps } from '.';
 import './index.css';
 import SignalRow from './signalRow';
+import TimelineRow from './TimelineRow';
 
-type Props = ReturnType<typeof mapStateToProps>;
+type Props = ReturnType<typeof mapStateToProps> & ReturnType<typeof mapDispatchToProps>;
 
-export default class SignalViewer extends React.PureComponent<Props> {
+function getClosestValue(signalArray, timeStamp, key) {
```
domoritz: Please use binary search to make this faster.

domoritz:
* [ ] The initial load in http://localhost:8080/#/examples/vega-lite/interactive_brush shouldn't create a bunch of entries.

Screen Shot 2019-08-21 at 12 31 34 AM

* [ ] Missing record toggle
* [ ] Errors in JS console
* [ ] Missing cell

Screen Shot 2019-08-21 at 12 32 11 AM

- [ ] Interacting with http://localhost:8080/#/examples/vega-lite/interactive_brush crashes the browser.

domoritz:
* [ ] Make the whole label "Enable timeline" clickable
* [ ] Change label to "Record signal changes"
* [ ] Use full width for signal viewer table when the timeline is on
* [ ] Why are undefined values not yellow?

Screen Shot 2019-08-25 at 10 40 31 AM

* [ ] "reset timeline" sounds scary. Maybe make it "Continue recording"
* [ ] Keep state of whether I am recording or not and don't reset when we change the spec. 
* [ ] Add a button to clear the timeline
* [ ] Consolidate changes. Changes that happen together should appear in the same column. 

Screen Shot 2019-08-25 at 10 43 33 AM

* [ ] Use full vertical space.

Screen Shot 2019-08-25 at 10 45 20 AM


* [ ] Add overview with zoom as in the original editor (or file an issue to do it later). See https://bl.ocks.org/mbostock/34f08d5e11952a80609169b7917d4172 for an example.

Screen Shot 2019-08-25 at 10 46 33 AM

package.json
```diff
@@ -16,19 +16,24 @@
     "prettier": "^1.18.2",
     "style-loader": "^1.0.0",
     "tslint": "^5.19.0",
-    "typescript": "~3.5.0",
-    "vega-lite-dev-config": "^0.2.5",
-    "webpack": "4.39.3",
+    "typescript": "~3.4.0",
```
domoritz: don't change these

siddhant1:
Can you check for problem now @domoritz

src/components/renderer/index.css
```diff
@@ -77,3 +78,18 @@
...
+
+.overlay {
+  position: absolute;
+  display: none;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background-color: rgba(238, 238, 239, 0.48);
+  background: url(data:;base64,iVBORw0KGgoAAAANSUhEUgAAAAIAAAACCAYAAABytg0kAAAAAXNSR0IArs4c6QAAAARnQU1BAACxjwv8YQUAAAAgY0hSTQAAeiYAAICEAAD6AAAAgOgAAHUwAADqYAAAOpgAABdwnLpRPAAAABl0RVh0U29mdHdhcmUAUGFpbnQuTkVUIHYzLjUuNUmK/OAAAAATSURBVBhXY2RgYNgHxGAAYuwDAA78AjwwRoQYAAAAAElFTkSuQmCC)
```
domoritz: What's this? Please add a comment.

src/components/renderer/index.css
```diff
@@ -1,6 +1,7 @@
 .chart {
   background: white;
   padding: 10px;
+  position: relative;
```
domoritz: Why change this?

package.json
```diff
@@ -24,11 +24,16 @@
...
   },
   "dependencies": {
+    "@types/d3": "^5.7.2",
     "@types/react": "^16.9.2",
     "@types/react-redux": "^7.1.2",
     "@types/react-router-dom": "^4.3.5",
     "ajv": "^6.10.2",
+    "d3-array": "^2.3.1",
     "d3-geo-projection": "^2.7.0",
+    "d3-request": "1.0.6",
```
domoritz: Please use fetch.

src/components/signal-viewer/TimelineRow.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export default class TimelineRow extends Component<
+  {
+    data: any[];
+    width: number;
+    xCount: number;
+    onHoverInit: (hoverValue: any) => void;
+    onHoverEnd: () => void;
+    onClickInit: (hoverValue: any) => void;
+    isClicked: boolean;
```
domoritz: What does it mean to be clicked? Please find a better name.

src/components/signal-viewer/renderer.tsx
```diff
@@ -1,54 +1,320 @@
...
+      return;
+    }
+    if (changeKey) {
+      const obj = {
+        value: this.props.view.signal(changeKey)
+      };
+      const lastObj = this.props.signals[changeKey];
+      const prevObj = {...lastObj[lastObj && lastObj.length - 1]};
+      delete prevObj.xCount;
+      // if (equalCycles(obj, prevObj)) {
```
domoritz: remove

src/components/signal-viewer/renderer.tsx
```diff
@@ -1,54 +1,320 @@
...
-        data: vega.falsy,
+      ref.view.getState({
+        data: vega.truthy,
         signals: vega.truthy,
         recurse: true
       }).signals
     );
   }
 
+  public getSignals(changeKey = null) {
```
domoritz: what does changeKey mean? Find a better name.

src/components/signal-viewer/renderer.tsx
```diff
@@ -1,54 +1,320 @@
...
+      },
+      () => {
+        setImmediate(() =>
+          this.setState({
+            maskListner: false
+          })
+        );
+      }
+    );
+    // remove the maskListner
```
domoritz: Why does this comment come afterward?

src/components/renderer/index.css
```diff
@@ -77,3 +78,18 @@
 .fullscreen-close:hover {
   background: rgba(0, 0, 0, 0.08);
 }
+
+.overlay {
```
domoritz: This selector is too generic.

src/components/signal-viewer/signalRow.tsx
```diff
@@ -26,22 +35,68 @@ export default class SignalRow extends React.PureComponent<Props, State> {
...
+    if (isClicked) {
+      return clickedSignal;
+    } else if (isHovered) {
+      return hoverValue;
+    } else {
+      return null;
+    }
+  };
+
+  public getBackgroundColor = () => {
```
domoritz: Please use css classes instead.

src/components/signal-viewer/renderer.tsx
```diff
@@ -1,54 +1,320 @@
...
-              />
-            ))}
-          </tbody>
-        </table>
-      </div>
+      <>
+        <div style={{display: 'inline-block'}}>
+          <button
+            className="sharing-button"
+            style={{backgroundColor: this.state.timeline ? 'red' : '', margin: '7px 10px'}}
```
domoritz: Please use a css class.

src/components/signal-viewer/renderer.tsx
```diff
@@ -1,54 +1,320 @@
...
+  }
+
+  public onHoverInit(signalKey, hoverValue, shouldPersist = false) {
+    const hoverObj = {
+      [signalKey]: hoverValue.value
+    };
+    const countObj = {
+      [signalKey]: hoverValue.xCount
+    };
+    Object.keys(this.props.signals).map(key => {
```
domoritz:
```
for (const key of keys()) {
}
```

import keys from vl.

src/components/signal-viewer/renderer.tsx
```diff
@@ -1,54 +1,320 @@
...
+      let i = 0;
+      while (this.props.signals[key][i] && this.props.signals[key][i].xCount <= hoverValue.xCount) {
```
domoritz: This still isn't using binary search but maybe that's okay.

domoritz:
@siddhant1 can you help finish this PR so we can release this feature to users?

siddhant1:
I think it's mostly complete , can you check for any changes in functionality

domoritz:
Sure. Can you rebase/merge master and then I will take a look and send you a review. There are still unresolved comments above. Can you make the ones that you fixed as resolved?

domoritz:
Please remove the WIP label when this is ready for a review.

siddhant1:
I have made most of the suggested changes, for the styles, I would like to keep them this way.
I have made some refactoring with the styles which were repetitive.

domoritz:
Thank you!

yarn.lock
```diff
@@ -2996,10 +3228,27 @@ d3-quadtree@1:
   resolved "https://registry.yarnpkg.com/d3-quadtree/-/d3-quadtree-1.0.6.tgz#d1ab2a95a7f27bbde88582c94166f6ae35f32056"
   integrity sha512-NUgeo9G+ENQCQ1LsRr2qJg3MQ4DJvxcDNCiohdJGHt5gRhBW6orIB5m5FJ9kK3HNL8g9F4ERVoBzcEwQBfXWVA==
 
+<<<<<<< HEAD
```
domoritz: Remove merge marker

domoritz:
* The red recording button has too low contrast. Can you just add a red cirle or make the text white?
* Stop recording clears the timeline. Should it?
* Let's reduce the opacity of the overlay.
* Add a tooltip to the overlay explaining how to get back to interacting with the chart. 
* related changes should be grouped together. Right now, we never have changes that belong together together. Maybe @jhoffswell has some suggestions for how to achieve grouping.

Screen Shot 2019-10-16 at 10 02 59

siddhant1:
Okay if we stop the recording on the button , what should happen to signal values , should they update?

domoritz:
I think we should pause recording and interactions similar to what happens when you select a previous timestamp. Does that make sense?

domoritz:
@siddhant1 Can you finish this? It would be great to get this in so people can use it.

siddhant1:
Fixed most of the issues , but not sure about the stop recording one , how will the user "stop the recording" then?

domoritz:
Thank you. Just a few small things left and then we can merge.

* Remove unsafe lifecycle methods similar to https://github.com/vega/editor/pull/537. 
* I removed some old classes and fixed the style of the buttons. Please review!
* If possible, please add some batching of the recorded changes. Instead of recording a every change, it would be nice to batch changes that occur within a short time (let's say 10ms). You can use throttling to achieve this.

siddhant1:
1. I have tried refactoring it a dozen of times but i am not able to figure out how to remove lifecycle hooks , i have tried every way but the app is breaking somehow.
2. Styles looks great
3. I think maybe we can hold on to the batching now , to add batching we have to change the entire structure , also i think that won't be helpful because throttling will result in some missing signals , also we will have to see how the signal changed and which 2 changed simultaneously.

domoritz:
Okay. Let’s fix the merge issues and I’ll take a look and merge later (I’m on vacation right now).

domoritz:
Okay, let's leave the lifecycle hooks for now. 

Let's try batching. At least events that happen within 1ms (or maybe 10ms) from each other should be considered as being caused by the same base event. @jhoffswell do you have a suggestion how we could handle batching?

domoritz:
@siddhant1 can you merge master into this? I'll make a pass and merge then.

domoritz:
🎉 

I think this can still be improved but let's merge it so we can get feedback.

PR #476, opened 2019-08-19
Use MessageData interface from vega-embed

danmarshall:


src/components/app.tsx
```diff
@@ -15,6 +16,14 @@ import InputPanel from './input-panel';
 import Sidebar from './sidebar';
 import VizPane from './viz-pane';
 
+interface MessageData {
```
domoritz: We should define this in Vega-Embed in https://github.com/vega/vega-embed/blob/master/src/post.ts#L10 and then import here.

danmarshall: Vega-embed is not yet a dependency... https://github.com/vega/editor/blob/master/package.json#L55

domoritz: Let's make it one then. Maybe there was a reason I didn't use it but I can't remember right now so let's try :-D It seems like the right thing to do.

danmarshall: Seems backwards... because Embed has the "Open in Vega Editor" button, it is the depend-er. Also, the editor seems to accept from any origin, not just the Embed.

domoritz: Embed is way smaller and more stable. Yes, the editor accepts post messages from other places but embed is the most well-defined and everyone else needs to adhere to its interface.

domoritz: Another way to put it is that I won't import the editor in embed ;-)

domoritz:
Thank you!

PR #478, opened 2019-08-20
Modify condition to check if user has gists

algomaster99:
### Fun fact
Even if `this.state.personalGist` is an empty array, `[] !== []` since they both reference a different memory location.

Therefore, it is better to check the length of `this.state.personalGist`.


domoritz:
Wohoo, Javascript.

PR #484, opened 2019-08-24
Switch to ESLint

domoritz:


siddhant1:
We need to add similar rules as before

domoritz:
@siddhant1 What kind of rules? The tests are passing.

algomaster99:
@domoritz I suggest keeping `editor-backend` and `editor` linting rules consistent. So you might want to copy the [configuration file in `editor-backend`](https://github.com/vega/editor-backend/blob/master/.eslintrc.json). Since this is a front-end project, we will need to add more react specific rules like this https://github.com/vega/editor/blob/1ec5cf5b6d0865aa9bf0da12a45444ca163e2bf3/package.json#L86 and they can simply be added on top.

domoritz:
Actually, we use a common config for all projects from https://github.com/vega/vega-lite-dev-config. Would you want to use that in the back end as well? If necessary, we can adjust the rules there as well. All I did in the package.json here is to override the defaults until we clean up the code.

Issue #49, opened 2018-03-13
Sharing visualizations


domoritz:
See https://docs.google.com/document/d/1YKBzT-aTrKGxA-G6ktzRSIC-ihc3JCWmNW5k7OL85nA/edit#heading=h.h1iauva9hmtp and look at https://twitter.com/mathisonian/status/973628667854123009 for an idea of what this could be like. 

@mathisonian might have some suggestions for how we can implement this.

mathisonian:
For Idyll's editor I use the following architecture:

`documentation website <-----> idyll api server <--------> amazon DynamoDB`

DynamoDB is a nice backend because it allows you to persist blobs of JSON without really worrying about too much beyond providing auth keys. 

The Idyll API server acts as a proxy to DynamoDB, allowing users the editor frontend to "save" by sending a POST request, and allowing the documentation site code to function without keys to the database. This provides some nice separation of concerns, and you can implement logic like rate limiting, etc on the api server. 

I'm using [now](https://zeit.co/now) to host the API server, its a nice managed environment for node.js code. I also power the documentation site with next.js and now, which gives a way to do dynamic URL routing with react apps, making URLs like https://idyll-lang.org/editor/5b2c7b11-22fb-4e69-98d9-de79a37e7f8d/view possible.

domoritz:
Thanks for the notes! I suppose for the storage any key-value store will work but it's great to hear that dynamo works well for this case. Can you share the server code?

mathisonian:
Yes - happy to share though its not cleaned up for github yet. Will post here once it is open source. The code itself is very simple though, essentially an express.js based proxy to the datastore

domoritz:
* [ ] Set up an oauth server with https://zeit.co/now
* [ ] Support anonymous sharing via spec in the URL
* [ ] Add login dialog
* [ ] Save to gist if the user is logged in
* [ ] Save specs to elastic so that we can search them

Code to look at https://github.com/enjalot/blockbuilder and https://github.com/vega/vega-editor/tree/saveInGist

domoritz:
Works now

Issue #448, opened 2019-08-12
Saving to GIST


algomaster99:
A button can be added inside export modal which will create a GitHub gist with the current vega/vega-lite specification written in the editor. (GSoC target)

Further, we can add forking and updating functionalities.

ashu8912:
Hey @algomaster99 can i takedown this issue

domoritz:
https://github.com/vega/editor/pull/489 is a start but needs to be finished.

domoritz:
Works now

Issue #371, opened 2019-07-05
Workflow of loading Gists


algomaster99:
This issue will discuss how the loading of gists will work.

Firstly, I am thinking of to divide the modal into two parts using a vertical divider. Something like this:
![Screenshot from 2019-07-05 16-07-00](https://user-images.githubusercontent.com/35191225/60717628-fe512e80-9f3f-11e9-85e2-77141ccccb6b.png)

#### Left
- The left portion will simply have input fields which will load any gist whose ID is available. (No need to enter the full URL). Can be achieved using [this reference](https://developer.github.com/v3/gists/#get-a-single-gist) from the guide.
- We can look for the `$schema` attribute in the content and find the visualization type. This will not enforce a user to add an extension to their gist.

#### Right
- If authenticated, it will show the list of all the gists the user has created. I have created the API for fetching the list. The response will look something like this:
![Screenshot from 2019-07-05 16-51-20](https://user-images.githubusercontent.com/35191225/60719402-28591f80-9f45-11e9-9d7d-94e4920663fa.png)
- This will be displayed as a list and upon clicking one of them, we can request the source code using the `raw_url`.
- It will further follow the same process to find the gist type.

algomaster99:
@domoritz Please suggest some ideas.

algomaster99:
### Mock-up of header
![Header](https://user-images.githubusercontent.com/35191225/60757596-5c5a4080-a02a-11e9-8eaf-20132ecfe45d.jpg)
The circular button on the leftmost side will show user profile.

### Load GIST
![Load GIST](https://user-images.githubusercontent.com/35191225/60757697-78aaad00-a02b-11e9-908c-cdeb948379f0.jpg)

### Save GIST
![Save to GIST](https://user-images.githubusercontent.com/35191225/60757700-82ccab80-a02b-11e9-9019-007feaee9d67.png)

Let me clarify the `File Name (auto-filled)` field. If the gist is already loaded from some file, its name will be filled there. Otherwise, you may change the name and chose the following two buttons. **Add a file** will add another file to the cuurent gist (make sure to rename the file name). **Edit the current file** will edit the content in the file or if the name is changed, it will be renamed.

### Figma Link
Please check [this link](https://www.figma.com/file/CoRMRbayr2OJ87aXlqmThd2Q/Vega-Header?node-id=0%3A1) for prototyping of the header. I have also created a design for **Save GIST** and **Load GIST** button which are attached above.

@domoritz @arvind

[...]

PR #489, opened 2019-08-30
Create gist

algomaster99:
This enables saving feature in the editor. Open the share modal and click create and it will directly create a **new** gist in your GitHub account if you are logged in.

- [ ] Feature of overriding a loaded gist.

Part of https://github.com/vega/editor/issues/49

### DIFF
--- a/src/components/header/share-modal/renderer.tsx
+++ b/src/components/header/share-modal/renderer.tsx
@@ -221,15 +221,17 @@ class ShareModal extends React.PureComponent<Props, State> {
               </a>
             </h3>
             <div className="share-input-container">
-              <input
-                type="checkbox"
-                name="private-gist"
-                id="private-gist"
-                value="private-select"
-                checked={this.state.gistPrivate}
-                onChange={this.updatePrivacy.bind(this)}
-              />
-              <label htmlFor="private-gist">Private gist</label>
+              <label>
+                <input
+                  type="checkbox"
+                  name="private-gist"
+                  id="private-gist"
+                  value="private-select"
+                  checked={this.state.gistPrivate}
+                  onChange={this.updatePrivacy.bind(this)}
+                />
+                Private Gist
+              </label>
             </div>
             <div className="share-input-container">
               <label>
--- a/src/components/header/share-modal/renderer.tsx
+++ b/src/components/header/share-modal/renderer.tsx
@@ -1,15 +1,14 @@
-import LZString from "lz-string";
-import * as React from "react";
-import Clipboard from "react-clipboard.js";
-import { Copy, Link } from "react-feather";
-import { withRouter } from "react-router-dom";
-import { mapStateToProps, mapDispatchToProps } from ".";
-import "./index.css";
-import getCookie from "../../../utils/getCookie";
-import { BACKEND_URL, COOKIE_NAME } from "../../../constants/consts";
+import LZString from 'lz-string';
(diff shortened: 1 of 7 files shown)


src/components/header/share-modal/renderer.tsx
```diff
@@ -63,55 +81,213 @@ class ShareModal extends React.PureComponent<Props, State> {
...
+      mode: "cors",
+      credentials: "include",
+      headers: {
+        "Content-Type": "application/json",
+        Cookie: `${COOKIE_NAME}=${cookieValue}`
+      },
+      method: "post"
+    })
+      .then(res => {
+        return res.json();
```
algomaster99: This line is throwing `SyntaxError: Unexpected token C in JSON at position 0`. Here is the answer I referred to solve this problem: https://stackoverflow.com/a/43094587/11751642. But this returns `{}` when user is not logged in which shouldn't be the case since it return `{"handle":"","isAuthenticated":false,"name":"","profilePicUrl":""}`
`

domoritz:
@algomaster99 is this ready? Can you help finish it so we can release it to users?

algomaster99:
@domoritz Will get to it as soon as possible. I have updated you on Slack.

algomaster99:
@domoritz do you have any idea about its implementation? Can you suggest me some links?

domoritz:
Since you already started an implementation, I don’t exactly understand what links you are looking for. Can you elaborate?

domoritz:
@algomaster99 Can you help get this into master?

algomaster99:
@domoritz I couldn't think of a workflow that will let me know if the gist being edited will be someone else's or our own. This could have been done keeping a state for it like `ownership` but I haven't thought it through.
Currently, you can just export it to a gist. Editing a gist is not available.

domoritz:
That’s okay for now. I think we can just let people edit and then when they try to save there will be an error. We can then offer to save as a forked gist.

domoritz:
Can you fix the build errors? @algomaster99

algomaster99:
@domoritz I think they are linting errors. Shall I install `prettier`?

domoritz:
`yarn format` should be enough. I just ran it and pushed.

src/components/header/share-modal/renderer.tsx
domoritz: Is this fixed? Sounds like we are not requesting a JSON.

algomaster99:
@domoritz Cool, will do it as soon as possible.

src/components/header/share-modal/renderer.tsx
algomaster99: @domoritz I couldn't exactly get why the error persisted though

ashu8912:
Hey @domoritz is this pr not yet complete i can help if its stuck.

domoritz:
@algomaster99 @ashu8912 Can you discuss who finishes this pull request?

algomaster99:
@ashu8912 Would you like to work on this?

ashu8912:
sure @algomaster99

algomaster99:
@ashu8912 https://github.com/vega/editor/issues/371 https://github.com/vega/editor/issues/448 refer to these.
In the first link, only the `saving to gist` has to be implemented. If you have anymore doubts, please feel to ping me on Slack.

algomaster99:
Continued in #691

PR #502, opened 2019-09-17
Fix mode switch bug

RisingGeek:
Vega and Vega-light modes were not switching repeatedly because the hash history was not able to push the same path.
Changing the route when the mode is switched to vega solved the issue.

### DIFF
--- a/src/components/header/renderer.tsx
+++ b/src/components/header/renderer.tsx
@@ -107,6 +107,10 @@ class Header extends React.PureComponent<Props, State> {
     this.props.history.push(`/examples/vega/${name}`);
   }
 
+  public onSelectNewVega() {
+    this.props.history.push('/');
+  }
+
   public onSelectVegaLite(name) {
     this.props.history.push(`/examples/vega-lite/${name}`);
   }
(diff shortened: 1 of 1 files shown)


domoritz:
Thank you! Shouldn't we switch to `/custom/vega`, though?

RisingGeek:
> Thank you! Shouldn't we switch to `/custom/vega`, though?

yaa I will change the route.

src/components/header/renderer.tsx
```diff
@@ -116,7 +120,9 @@ class Header extends React.PureComponent<Props, State> {
   }
 
   public onSwitchMode(option) {
-    option.value === Mode.Vega ? this.props.updateVegaSpec(stringify(this.props.vegaSpec)) : this.onSelectNewVegaLite();
```
domoritz: Can you replace this with an if block instead of a ternary?

RisingGeek: yes, I will replace it with if-else block

src/components/header/renderer.tsx
```diff
@@ -116,7 +120,11 @@ class Header extends React.PureComponent<Props, State> {
...
+    if (option.value === Mode.Vega) {
+      this.props.updateVegaSpec(stringify(this.props.vegaSpec)) && this.onSelectNewVega();
```
domoritz: Should this also be an if or just two commands?

src/components/header/renderer.tsx
domoritz: Thanks

src/components/header/renderer.tsx
RisingGeek: We can put it in an if condition. The current state only shortens the code. Should I make another if?

domoritz: It's much clearer, yes.

RisingGeek: Oh! there was no need for if condition. The route will change every time. My bad

domoritz: I'm glad clarifying the code helped you as well.

PR #503, opened 2019-09-18
Redirect back after logging in from gist modal

algomaster99:
One possible solution can be to simply remove `rel='noopener noreferrer'` from `a` tags but that is prone to [security threats](https://mathiasbynens.github.io/rel-noopener/) and [eslint](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md) won't allow us to do it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,6 @@
         "@typescript-eslint/no-var-requires": "off",
         "react/no-find-dom-node": "warn",
         "react/no-unescaped-entities": "off",
-        "react/jsx-no-target-blank": "off",
         "@typescript-eslint/no-unused-vars": "warn",
         "no-prototype-builtins": "warn"
       }
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
         "@typescript-eslint/no-var-requires": "off",
         "react/no-find-dom-node": "warn",
         "react/no-unescaped-entities": "off",
-        "react/jsx-no-target-blank": "warn",
+        "react/jsx-no-target-blank": "off",
         "@typescript-eslint/no-unused-vars": "warn",
         "no-prototype-builtins": "warn"
       }
--- a/package.json
+++ b/package.json
@@ -88,6 +88,7 @@
         "@typescript-eslint/no-var-requires": "off",
         "react/no-find-dom-node": "warn",
         "react/no-unescaped-entities": "off",
+        "react/jsx-no-target-blank": "warn",
         "@typescript-eslint/no-unused-vars": "warn",
         "no-prototype-builtins": "warn"
       }
--- a/src/components/header/gist-modal/renderer.tsx
+++ b/src/components/header/gist-modal/renderer.tsx
@@ -417,7 +417,7 @@ class GistModal extends React.PureComponent<Props, State> {
               )
             ) : (
               <span>
-                Login with{' '}
+                Login with {/* eslint-disable-next-line react/jsx-no-target-blank */}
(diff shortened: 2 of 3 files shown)


domoritz:
Don't we open a new tab and then close it? 

If necessary, we can disable the check for this line since we control the backend service and trust it.

algomaster99:
We open the new-tab and then close it. But we need the opener as well since an event listener is attached to it. When `rel='noopener'` is used it passes `window.opener` as null.

package.json
```diff
@@ -88,6 +88,7 @@
         "@typescript-eslint/no-var-requires": "off",
         "react/no-find-dom-node": "warn",
         "react/no-unescaped-entities": "off",
+        "react/jsx-no-target-blank": "warn",
```
domoritz: Instead of this, let's silence the warning in the one case below only.

algomaster99: @domoritz Do you mean to change it to 'off'?

domoritz: No, use `// eslint-disable-next-line react/jsx-no-target-blank`

src/components/header/gist-modal/renderer.tsx
```diff
@@ -417,7 +417,7 @@ class GistModal extends React.PureComponent<Props, State> {
               )
             ) : (
               <span>
-                Login with{' '}
+                Login with {/* eslint-disable-next-line react/jsx-no-target-blank */}
                 <a href={`${BACKEND_URL}auth/github`} target="_blank">
```
domoritz: Let's use `/* eslint-disable-line react/jsx-no-target-blank */` on the actual line.

algomaster99: @domoritz is this what you mean?
```
{/* eslint-disable-next-line react/jsx-no-target-blank */} <a href={`${BACKEND_URL}auth/github`} target="_blank"> 
```
ESLint automatically lints it to the following
```
Login with {/* eslint-disable-next-line react/jsx-no-target-blank */}
<a href={`${BACKEND_URL}auth/github`} target="_blank">
```

domoritz: Disable lint for the actual line, not the next line.

algomaster99: @domoritz So I used `// eslint-disable-line react/jsx-no-target-blank`.
```
<span>
  Login with{' '}
  <a href={`${BACKEND_URL}auth/github`} target="_blank"> {/* eslint-disable-line react/jsx-no-target-blank */}
    GitHub
  </a>{' '}
  to see all of your personal gist.
</span>
```
But gets transformed to:
```
<span>
  Login with{' '}
  <a href={`${BACKEND_URL}auth/github`} target="_blank">
    {' '}
    {/* eslint-disable-line react/jsx-no-target-blank */}
    GitHub
  </a>{' '}
  to see all of your personal gist.
</span>
```
and fails `yarn lint`.

domoritz: Ok

domoritz: How about you extract the link into a variable then.

algomaster99: No, it doesn't work

algomaster99:
```
/* eslint-disable */
<a href={LOGIN_URL} target="_blank">
/* eslint-enable */
```
How about this?

domoritz: I'll play with it and merge.

Issue #443, opened 2019-08-07
Lazy loading code


domoritz:
We should consider using https://reactjs.org/docs/code-splitting.html#reactlazy for code that is not always needed like the sharing via URL. This could reduce the startup time of the editor.

siddhant1:
Lets do it!

domoritz:
We need to be careful and profile whether this makes a difference.

domoritz:
Let's not do this until we have proof that code splitting actually helps.

PR #506, opened 2019-09-29
Lazy loading for header components

RisingGeek:
Lazy loading code - #443 
- Changed module to esnext in tsconfig(for lazy loading to work with typescript)

- Added lazy loading to export, share, gist, help components.

### DIFF
--- a/src/components/header/renderer.tsx
+++ b/src/components/header/renderer.tsx
@@ -10,8 +10,8 @@ import {BACKEND_URL, COOKIE_NAME, KEYCODES, Mode} from '../../constants';
 import {NAMES} from '../../constants/consts';
 import {VEGA_LITE_SPECS, VEGA_SPECS} from '../../constants/specs';
 import getCookie from '../../utils/getCookie';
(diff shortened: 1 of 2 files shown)


domoritz:
Thank you for the PR. Did you check how much smaller the bundles are now and whether the website loads faster?

RisingGeek:
No. What is the process to check the bundle size?

domoritz:
Check out the analyze command in the package.json. For loading, you can use throttling in the dev tools of your browser.

RisingGeek:
The bundle size increased when I changed module to esnext. But with es2015, dynamic imports are not supported.

domoritz:
That's okay. I just wonder whether it's worth using lazy loading at all then.

RisingGeek:
There were many issues related to dynamic imports and the solution given to that was changing the module to esnext or commonjs in tsconfig. I will research more on this if there is any other solution.

domoritz:
Thanks. I think I'll close this PR until we know that there is real benefit of using lazy loading.

PR #521, opened 2019-10-02
Maximise container width for personal gists

### LINKED ISSUE #520: Fix size of gist list container
To make the size more consistent, let's resize the container with the list in it to take the full width and more height (look how much space is wasted in my screenshot).

Screen Shot 2019-09-30 at 23 47 49

algomaster99:
Fixes #520

### DIFF
--- a/src/components/header/gist-modal/index.css
+++ b/src/components/header/gist-modal/index.css
@@ -21,7 +21,7 @@
 }
 
 .gist-wrapper {
-  max-width: 35%;
+  margin-right: 0.5em;
   overflow-y: auto;
   max-height: 35vh;
 }
--- a/src/components/header/gist-modal/index.css
+++ b/src/components/header/gist-modal/index.css
@@ -1,8 +1,10 @@
 .gist-split {
   display: flex;
+  flex-wrap: wrap;
 }
 
 .personal-gist {
+  min-width: 50%;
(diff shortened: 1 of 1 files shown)


domoritz:
Thanks!

We're still not using the available height when we could.
Screen Shot 2019-10-02 at 09 50 29


Also, on small screens, the layout looks very odd. Can you improve it?

Screen Shot 2019-10-02 at 09 50 22

algomaster99:
@domoritz 
wrap

domoritz:
Does the right side of the popup need to be this wide?

algomaster99:
@domoritz I think the widths of input fields are fixed. Will have to make it flexible.

domoritz:
Let's do that. It would be nice if we split the width in half and at some width move the list above the form as you've shown.

algomaster99:
@domoritz done

domoritz:
Looks good. Now that I've used the gist loading a but more, I think we should fade out the list while we load another one instead of showing "Loading..." to avoid jumps in the UI. What do you think?

algomaster99:
@domoritz like fade out the old one and fade in the new?

domoritz:
I think just while the new list is loading, reduce the opacity of the old list. Nothing facny.

PR #522, opened 2019-10-02
Use async-await instead of promise chain

### LINKED ISSUE #483: Add error handling
We probably don't want to ignore errors here:

https://github.com/vega/editor/blob/1b2cbcf4d48bbf6871e16787d2fe51dae3ae20b8/src/components/header/renderer.tsx#L79

While you're at it, can you replace the code with async?

algomaster99:
Fixes #483

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -115,23 +115,27 @@ class App extends React.PureComponent<Props & {match: any; location: any; showEx
   }
 
   public async setGist(parameter: {id: string; filename: string; revision?: string}) {
-    await fetch(
-      `https://api.github.com/gists/${parameter.id}${parameter.revision !== undefined ? `/${parameter.revision}` : ''}`
-    )
-      .then(res => res.json())
-      .then(json => {
-        const contentObj = JSON.parse(json.files[parameter.filename].content);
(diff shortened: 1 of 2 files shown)


algomaster99:
@domoritz 
I just haven't converted [this](https://github.com/vega/editor/blob/5d6924e2b24cfc14f1a6e92a81b3c7668a8561ad/src/components/header/gist-modal/renderer.tsx#L136) because it is wired up like that. I was having difficulty to do it without a promise chain.

src/components/app.tsx
```diff
@@ -115,23 +115,27 @@ class App extends React.PureComponent<Props & {match: any; location: any; showEx
...
-          } else if (mode === Mode.VegaLite) {
-            this.props.setGistVegaLiteSpec('', json.files[parameter.filename].content);
-          }
+    try {
+      const response = await fetch(
+        `https://api.github.com/gists/${parameter.id}${
+          parameter.revision !== undefined ? `/${parameter.revision}` : ''
+        }`
+      );
+      const data = await response.json();
```
domoritz:
```suggestion
      const data = await response.json();
      const content = data.files[parameter.filename].content;
```

PR #523, opened 2019-10-03
Add os upgrade command into Dockerfile

y-vectorfield:
I tried to build docker image file many times, however the step of rsync installation was terminated with error. If we would like to complete building docker image for Vega editor, we need to update the os of the container image. Hence, I added this modification.


y-vectorfield:
Thank you very much for merging my commit into master.

domoritz:
Thank you for your contribution.

PR #525, opened 2019-10-04
feat: add resize handle for responsive charts

donghaoren:
Add a resize handle for responsive charts. The resize handle appears when the chart is supposed to be responsive, precisely:
- Either width or height is defined as a signal, and
- the signal uses `containerSize()[0]` or `containerSize()[1]` as its `init`.

### DIFF
--- a/src/components/renderer/renderer.tsx
+++ b/src/components/renderer/renderer.tsx
@@ -57,37 +57,36 @@ class Editor extends React.PureComponent<Props, State> {
     }
   }
 
-  public isResponsive(): [boolean, boolean] {
+  public isResponsive() {
     const spec = this.props.vegaSpec;
-    let w = false;
-    let h = false;
-    if (spec.signals && spec.signals instanceof Array) {
+    let responsiveWidth = false;
+    let responsiveHeight = false;
+    if (spec.signals) {
       for (const signal of spec.signals) {
-        if (signal.name == 'width' && (signal as vega.InitSignal).init == 'containerSize()[0]') {
-          w = true;
+        if (signal.name == 'width' && (signal as vega.InitSignal).init.indexOf('containerSize') >= 0) {
+          responsiveWidth = true;
         }
-        if (signal.name == 'height' && (signal as vega.InitSignal).init == 'containerSize()[1]') {
-          h = true;
+        if (signal.name == 'height' && (signal as vega.InitSignal).init.indexOf('containerSize') >= 0) {
+          responsiveHeight = true;
         }
       }
     }
-    return [w, h];
+    return {responsiveWidth, responsiveHeight};
   }
 
   public handleResizeMouseDown(e: React.MouseEvent) {
     const x0 = e.pageX;
     const y0 = e.pageY;
-    const width0 = this.state.width;
-    const height0 = this.state.height;
-    const [resizeWidth, resizeHeight] = this.isResponsive();
+    const {width: width0, height: height0} = this.state;
+    const {responsiveWidth, responsiveHeight} = this.isResponsive();
     const onMove = (eMove: MouseEvent) => {
       const x1 = eMove.pageX;
       const y1 = eMove.pageY;
       const factor = this.state.fullscreen ? 2 : 1;
       this.setState(
         {
-          width: resizeWidth ? Math.max(10, width0 + (x1 - x0) * factor) : this.state.width,
-          height: resizeHeight ? Math.max(10, height0 + (y1 - y0) * factor) : this.state.height
+          width: responsiveWidth ? Math.max(10, width0 + (x1 - x0) * factor) : this.state.width,
+          height: responsiveHeight ? Math.max(10, height0 + (y1 - y0) * factor) : this.state.height
         },
         () => {
           // Dispatch window.resize, that currently the only way to inform Vega about container size change.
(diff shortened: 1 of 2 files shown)


src/components/renderer/renderer.tsx
```diff
@@ -57,6 +57,52 @@ class Editor extends React.PureComponent<Props, State> {
...
+      for (const signal of spec.signals) {
+        if (signal.name == 'width' && (signal as vega.InitSignal).init == 'containerSize()[0]') {
+          w = true;
+        }
+        if (signal.name == 'height' && (signal as vega.InitSignal).init == 'containerSize()[1]') {
+          h = true;
+        }
+      }
+    }
+    return [w, h];
```
domoritz: Return an object instead.

src/components/renderer/renderer.tsx
```diff
@@ -57,6 +57,52 @@ class Editor extends React.PureComponent<Props, State> {
...
+  }
+
+  public handleResizeMouseDown(e: React.MouseEvent) {
+    const x0 = e.pageX;
+    const y0 = e.pageY;
+    const width0 = this.state.width;
+    const height0 = this.state.height;
+    const [resizeWidth, resizeHeight] = this.isResponsive();
```
domoritz: Use destructor.

src/components/renderer/renderer.tsx
```diff
@@ -57,6 +57,52 @@ class Editor extends React.PureComponent<Props, State> {
     }
   }
 
+  public isResponsive(): [boolean, boolean] {
```
domoritz: add a comment about what this does and what the assumptions are

src/components/renderer/renderer.tsx
```diff
@@ -57,6 +57,52 @@ class Editor extends React.PureComponent<Props, State> {
...
+    const spec = this.props.vegaSpec;
+    let w = false;
+    let h = false;
+    if (spec.signals && spec.signals instanceof Array) {
```
domoritz:
```suggestion
    if (spec.signals) {
```

src/components/renderer/renderer.tsx
```diff
@@ -121,6 +170,19 @@ class Editor extends React.PureComponent<Props, State> {
       .renderer(this.props.renderer)
       .initialize(chart)
       .runAsync();
+
+    // Trigger the resize event if the chart is responsive. This seem to be necessary.
```
domoritz: Let's remove this. Vega should handle this properly, imho. We can leave this for now but file a vega issue and link to it from here.

src/components/renderer/renderer.tsx
```diff
@@ -57,37 +57,36 @@ class Editor extends React.PureComponent<Props, State> {
...
+    return {responsiveWidth, responsiveHeight};
   }
 
   public handleResizeMouseDown(e: React.MouseEvent) {
     const x0 = e.pageX;
     const y0 = e.pageY;
-    const width0 = this.state.width;
-    const height0 = this.state.height;
-    const [resizeWidth, resizeHeight] = this.isResponsive();
+    const {width: width0, height: height0} = this.state;
```
domoritz: Do you want to leave this as it was or also change the `x0`, `y0` above and `x1`, `y1` below?

donghaoren: I'd prefer not to rename in destructor. Reverted back, also added a few more comments to the code. I think it's good to go. Do you prefer squash merge or regular merge?

domoritz:
Thank you @donghaoren

src/components/renderer/renderer.tsx
donghaoren: @domoritz, I was about to open an issue in Vega, but with some further investigation, I found that the problem is because we are calling `vegaTooltip.default(view)` before `view.renderer` and `view.initialize`. `vegaTooltip.default` seem to call `view.run` (https://github.com/vega/vega-tooltip/blob/master/src/index.ts), that's why the signals got initialized before we do `view.initialize`

The following code:
```html
<!DOCTYPE html>
<script src="https://cdn.jsdelivr.net/npm/vega@latest"></script>
<script src="https://cdn.jsdelivr.net/npm/vega-tooltip@latest"></script>
<div id="view" style="width: 500px; height: 500px"></div>
<script>
  const spec = {
    $schema: "https://vega.github.io/schema/vega/v5.json",
    signals: [
      {
        name: "width",
        init: "warn(containerSize()[0])",
        on: [{ events: "window:resize", update: "containerSize()[0]" }]
      },
      {
        name: "height",
        init: "warn(containerSize()[1])",
        on: [{ events: "window:resize", update: "containerSize()[1]" }]
      }
    ],
    background: "yellow"
  };

  // editor's initView()
  const view = new vega.View(vega.parse(spec), {
    logLevel: vega.Warn,
    hover: true // enable hover processing
  }).hover();

  console.log("before tooltip");
  vegaTooltip.default(view);
  console.log("after tooltip");

  // editor's renderVega()
  view.renderer("canvas");
  view.initialize(document.getElementById("view"));
  view.runAsync();
</script>
```

I think there is another way to call tooltip. Let me try and I'll send another PR.

domoritz: Nice find. Thanks for looking into this!

PR #543, opened 2019-10-22
fixed feedback links clickable throughout the modal width

nik72619c:
@domoritz please have a look and suggest changes if any :)

### DIFF
--- a/src/components/header/help-modal/index.css
+++ b/src/components/header/help-modal/index.css
@@ -15,7 +15,7 @@
 }
 
 .help-modal .doc-link {
-  display: block;
+  display: inline-block;
   margin-bottom: 8px;
 }
 
--- a/src/components/header/help-modal/index.tsx
+++ b/src/components/header/help-modal/index.tsx
@@ -52,15 +52,23 @@ const HelpModal = () => {
         <AlertCircle className="header-icon" />
         <span>Report a Bug</span>
       </a>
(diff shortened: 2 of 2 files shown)


src/components/header/help-modal/index.css
```diff
@@ -15,7 +15,7 @@
 }
 
 .help-modal .doc-link {
-  display: block;
+  display: inline-block;
```
domoritz: Is this necessary with the new div?

nik72619c: yes it is. display: block is again gonna make the anchor tag utilise the whole of the width again

domoritz: Can you just delete the line?

domoritz: Links are by default inline. 

The other option is to not use the div and just set the display of the link.

siddhant1: Make the anchor tags inline-block , anchor tags are inline and won't accept any styles , we only need to change the style from block to inline-block , no need for adding <div> tags

siddhant1: This is an interesting one actually, if we set to inline-block we will get 2 links side by side , we need to set them to a inline-block and break lines after each ```<a>``` , so the solution can be to use ```<br/>``` or using inline-table or use this https://css-tricks.com/injecting-line-break/

PR #544, opened 2019-10-22
navigate to edited on merge config spec

### LINKED ISSUE #529: merging the config into the spec should navigate to /edited
Right now, we stay on the same page, which may be an example page.

shakti97:
Fix #529


src/components/config-editor/renderer.tsx
```diff
@@ -25,6 +28,9 @@ export default class ConfigEditor extends React.PureComponent<Props> {
     if (!confirmation) {
       return;
     }
+    if (this.props.history.location.pathname.indexOf('/edited') === -1) {
```
siddhant1: I think we can just check for '/edited' instead of linear search

domoritz:
@shakti97 @siddhant1 can you finish this?

shakti97:
Done with the changes recommended by @siddhant1 , @domoritz can you please check it and if any changes needed, let me know :)

domoritz:
Thank you for the fix. I wonder whether we can not write to the history since navigating back doesn't actually restore a previous state. What do you think?

domoritz:
Thank you for the fix. I wonder whether we can not write to the history since navigating back doesn't actually restore a previous state. What do you think?

PR #558, opened 2019-12-05
Refactor CSS of components

IamRavikantSingh:
For verification:
Are these changes correct as per requirements?
If OK then I can apply these changes to all other components.
@domoritz

### DIFF
--- a/src/components/app.css
+++ b/src/components/app.css
@@ -35,6 +35,7 @@
   --url-color: rgb(6, 69, 173);
   --box-shadow-color: rgba(0, 0, 0, 0.2);
   --transparent-color: rgba(0, 0, 0, 0.08);
+  --transparent-bg-color: rgba(0, 0, 0, 0.8);
 }
 
 body,
--- a/src/components/app.css
+++ b/src/components/app.css
@@ -29,6 +29,12 @@
   --dark-brand-color: rgb(92, 64, 141);
   --error-color: #f44336;
   --warning-color: #ffc107;
+  --box-shadow-color: rgba(0, 0, 0, 0.2);
+  --kbd-color: #444d56;
+  --kbd-bg-color: #fafbfc;
+  --kbd-border-color: #c6cbd1;
+  --kbd-border-bottom-color: #959da5;
+  --kbd-box-shadow-color: #959da5;
 
   --data-viewer-header-btn-bg-color: white;
   --data-viewer-header-btn-border-color: #ccc;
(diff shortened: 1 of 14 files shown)


domoritz:
Thanks for checking. I’d try to reduce the number of colors. It’s okay to change the look a bit as well. 

My goal with any refactoring is simplicity and consistency. 

What do you think if we just had a small set of colors (shades of gray and a few colors) and used them throughout the css?

IamRavikantSingh:
> Thanks for checking. I’d try to reduce the number of colors. It’s okay to change the look a bit as well.
> 
> My goal with any refactoring is simplicity and consistency.
> 
> What do you think if we just had a small set of colors (shades of gray and a few colors) and used them throughout the css?

So I have to eliminate colours which are very close?
One question. Should I keep the colour variable as it is or change it also I mean for easy control of colours of components we should keep them separate so that changing colour of one component should not affect others and then we can do is define set of colours and using it with those variables.

domoritz:
I'd say if we use colors that don't have meaning (like border color) and instead just name the variables after the color (red, lightgray), we don't have this issue.

IamRavikantSingh:
@domoritz please check the new commit.

domoritz:
Somehow the modal darkening is gone. 
Screen Shot 2019-12-07 at 14 53 44

The bg is a bit too dark now.

Screen Shot 2019-12-07 at 14 54 12

Also, it should probably be the same as this bg

Screen Shot 2019-12-07 at 14 55 01

Other than that, awesome work! I'm looking forward to merging this. 🎉

IamRavikantSingh:
I am not sure what to do with this colour so I left it as it is.
.myInlineDecoration {
  background-color: #f8c9aa !important;
}
in viz pane/css

src/components/app.css
```diff
@@ -21,14 +21,21 @@
 
   /* Colors */
   --light-color: #dedede;
-  --dark-color: #24292e;
+  --dark-color: #1a1d20;
   --primary-color: #f5f5f5;
```
domoritz: Calling this the primary colors is a bit odd. Usually, the primary color is a color like red or purple, not a shade of gray. Maybe let's rename this to darkgray, lightgray, etc. I know this was something you didn't change but I think it would be nice. 

Let's do this in a follow up. I'll merge for now.

domoritz:
> myInlineDecoration

Can you delete it? Where is it even used? Can you check?

domoritz:
Thank you for the pull request @IamRavikantSingh! I merged it and we will publish it with the next release.

domoritz:
I hope you see more opportunities for refactoring as you look through the code. Feel free to send anything as prs.

IamRavikantSingh:
> > myInlineDecoration
> 
> Can you delete it? Where is it even used? Can you check?

It is used in viz pane/renderer.tsx. I am not sure what it does there.
I am new to code so I need some time to read it. 😅
 ```
const decorations = editor.deltaDecorations(
      [],
      rangeValue.map(match => {
        return {
          options: {inlineClassName: 'myInlineDecoration'},
          range: match.range
        };
      })
    );
```

File: src/components/app.css (hunk 1 above)
Comment (IamRavikantSingh) [reply src/components/app.css]: Ok, I can fix it 👍

**Comment by siddhant1** [comment]
Myinlinedecoration is used to highlight instances of variable when the signal/data is clicked , we can probably rename it

File: src/components/app.css (hunk 1 above)
Comment (IamRavikantSingh) [reply src/components/app.css]:
```
  --white-color: #ffffff;
  /* --primary-light-color: #ffffff; */

  --light-gray-color: #f5f5f5;
  /* --primary-color: #f5f5f5; */

  --gray-color: #dedede;
  /* --light-color: #dedede; */

  --dark-gray-color: #c0c0c0;
  /* --primary-dark-color: #c0c0c0; */

  --extra-dark-gray-color: #666;
  /* --extra-dark-gray-color: #666; */
```

what about these names? any suggestions for naming.

domoritz:
And add a comment.

PR #581, opened 2019-12-29
Fixed Issue with cursor pointer

### LINKED ISSUE #546: Data Viewer and Signal Viewer table body Cursor Pointer
Showing Cursor Pointer without any clickable action on the table body of data viewer and signal viewer

PunitLodha:
fixed #546 by giving a pointer class to all clickable actions


domoritz:
Thank you

PR #591, opened 2020-01-24
Basic testing setup (Demo)

domoritz:
Just to demonstrate how to get started. Not working yet.

@IamRavikantSingh

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ _site
 .eslintrc.js
 .idea
 .prettierignore
+babel.config.js
 beemo-error.log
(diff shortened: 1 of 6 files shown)


siddhant1:
for basic testing setup this is fine , we probally need a renderer and a test runner to make our lives easy.
I am thinking of rtl or enzyme.
rtl doesn't do shallow so it can be slow sometimes for huge trees , other than that I am in favour of it since its very close to how the user uses the application

domoritz:
Lets actually make this work first, though.

PR #595, opened 2020-01-29
refactor-css config header

PunitLodha:
old:- 
Screenshot_2020-01-29 Editor IDE for Vega and Vega-Lite

new:-
Screenshot_2020-01-30 Editor IDE for Vega and Vega-Lite

### DIFF
--- a/src/components/config-editor/config-editor.css
+++ b/src/components/config-editor/config-editor.css
@@ -1,7 +1,3 @@
-.config-header {
-  text-transform: none;
-}
-
 .config-header > select {
   margin-left: 6px;
 }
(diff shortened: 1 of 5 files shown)


domoritz:
What problem is this solving?

PunitLodha:
The button was getting cut-off near the end. Also the text-transform: Uppercase property was applied to the whole config-header, which lead to applying another text-transform property on the theme label to make it lowercase.Fixed that by applying the text-transform property to only the tabs part

domoritz:
Thank you. I have to check this on a different OS and will give you feedback.

domoritz:
Hmm, I think the bar is now a bit too large and there are visual inconsistencies now. Maybe we can switch to the react select so that the style across OSes is the same. 

Screen Shot 2020-01-30 at 09 40 37
Screen Shot 2020-01-30 at 09 40 05
Screen Shot 2020-01-30 at 09 40 01

PR #596, opened 2020-01-30
Use BEM

PunitLodha:
Refactored CSS to use BEM. @domoritz please review this, so that i can start working on the other components

Also, how can i add a WIP label to this?

### DIFF
--- a/src/components/error-pane/index.css
+++ b/src/components/error-pane/index.css
@@ -1,12 +1,12 @@
-.error-pane .error {
+.error-pane__list-item--error {
   color: var(--error-color);
 }
 
-.error-pane .warning {
+.error-pane__list-item--warning {
   color: var(--warning-color);
 }
 
-.error-pane .info {
+.error-pane__list-item--info {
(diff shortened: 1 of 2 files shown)


domoritz:
Thank you, I added the WIP label.

src/components/error-pane/index.css
```diff
@@ -1,12 +1,12 @@
-.error-pane .error {
+.error-pane__list-item--error {
```
domoritz: I'm not a fan of long CSS names. If you have enough time, maybe it's worth rewriting everything to be more semantic with something like tachyons.

PunitLodha: Yes, BEM class names tend to be quite long.
So should i avoid using BEM and use tachyons classes instead?

domoritz: I've actually never used BEM and I can't say I'm convinced by its value yet. Let me close this pull request since I don't want to adopt it now.

PR #597, opened 2020-01-31
Rearrange and clean up Dockerfile commands for it to work on windows

abzachshan:
COPY overwites the modified scripts files so I moved the removal of 'r' characters AFTER the copy

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,22 +10,18 @@ RUN apt-get update && apt-get upgrade -y
 # Install rsync as it is a dependency of ./scripts/vendor.sh
 RUN apt-get -y install rsync
 
+# Install dos2unix to remove 'r' characters from scripts
+RUN apt-get install dos2unix
+
 # Sets the working directory for any RUN, CMD, ENTRYPOINT, COPY and ADD instructions that follow it in the Dockerfile
 # https://docs.docker.com/engine/reference/builder/#workdir
 WORKDIR /usr/src/app
 
-# Copies the package.json and yarn.lock files first to ensure the cache is only invalidated when these files change
-# https://nodejs.org/en/docs/guides/nodejs-docker-webapp/#creating-a-dockerfile
-COPY package.json yarn.lock ./
-
-# For this project, additional files must also be copied as yarn hooks depend on them
-COPY scripts ./scripts
-
-# Remove 'r' characters from the vendor script (otherwise it won't execute)
-RUN sed $'s/\r$//' ./scripts/vendor.sh > ./scripts/vendor.sh
-
-# Copy remaining files
+# Copy all files
 COPY . .
+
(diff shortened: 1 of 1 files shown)


domoritz:
Does this work even if I have a mode modules folder locally already? Does this automatically update when I change JS or CSS files? Please test this change carefully and report the results here. Thanks!

abzachshan:
> Does this work even if I have a mode modules folder locally already?

I'll check the node_modules folder - what's the expected behavior? Should it ignore the the exist folder? should we not copy the local node_modules to the container so it'll always be a fresh install?

>  Does this automatically update when I change JS or CSS files?

I'm not sure how it works on Linux/MacOS but in windows you'll have to build it again after changes to run include the newest changes (docker-compose build + up). Or you could change the files from the container itself...

Thanks!

domoritz:
Before, I think it would pick up changes automatically and reload the editor.

baumanab:
It works, thank you!  I also created this.  It has a different start image, command chaining, and some install cleanup.  I'm also using docker run in powershell instead of the compose file.  Docker run gives me a little more flexibility and powershell is kinder with mounts than git bash, when using a windows host.

`docker build -t vle .`

`docker run -dit -p 8080:8080 -v ${PWD}/src:/usr/src/app/src vle`

```
FROM node:latest

# Fetch updated list of packages and upgrade operating system for next step (install rsync)
RUN apt-get update && apt-get upgrade -y && \
    apt-get -y install rsync \
      dos2unix && \
    rm -rf /tmp/downloaded_packages/ /tmp/*.rds && \
    rm -rf /var/lib/apt/lists/*

# Sets the working directory for any RUN, CMD, ENTRYPOINT, COPY and ADD instructions that follow it in the Dockerfile
# https://docs.docker.com/engine/reference/builder/#workdir
WORKDIR /usr/src/app

# Copy remaining files
COPY . .

# Remove 'r' characters from the vendor script (otherwise it won't execute)
RUN dos2unix ./scripts/vendor.sh \
    ./scripts/version.sh  \
    ./scripts/generate-example-images.sh

# Run Yarn
RUN yarn

# Sets the container executable (ENTRYPOINT) as yarn and the default argument (CMD) as start
# https://docs.docker.com/engine/reference/builder/#entrypoint
# https://docs.docker.com/engine/reference/builder/#cmd
ENTRYPOINT ["yarn"]
CMD ["start"]
```

baumanab:
I also mounted a local folder to explore my own data.  ``` docker run -dit -p 8080:8080 -v ${PWD}/src:/usr/src/app/src -v  ${PWD}/data:/usr/src/app/public/data_local vle``` @domoritz @abzachshan would anything I've done be handy for the README or other documentation.  If so I can add it in and make a PR for either additional documentation or part of the README.

domoritz:
Thanks. Merged manually.

Issue #602, opened 2020-02-11
Config Editor Transferring Focus To Vega Editor


ashu8912:
Steps To Reproduce This Bug
1.Change Something In the config.
2.Press Run 
3.Now Start Changing then Config again 
4.Anything you type now is typed in the vega editor and not in the config editor (The Config Editor has Lost Focus somehow and transferred to vega editor).

domoritz:
Thanks for the bug report. Can you send a fix?

ashu8912:
Yes @domoritz Will Send a pr that fixes this

PR #603, opened 2020-02-13
fixed focus bug on config

ashu8912:
solves #602


ashu8912:
@domoritz

domoritz:
Thank you for the fix @ashu8912!

PR #617, opened 2020-02-18
feat: added warning for version mismatch

### LINKED ISSUE #447: Warn about incompatible versions
We should warn users about versions mismatches like Vega-Lite does: https://github.com/vega/vega-embed.

adrijshikhar:
Fix #447 
 * added warning for mismatch of current vega and vegaLite with the spec version input by user
screen-cast-2020-02-19_02 04 42

### DIFF
--- a/src/reducers/index.ts
+++ b/src/reducers/index.ts
@@ -1,5 +1,7 @@
 import stringify from 'json-stringify-pretty-compact';
-import {mergeConfig} from 'vega';
+import {mergeConfig, version} from 'vega';
+import {satisfies} from 'semver';
+import schemaParser from 'vega-schema-url-parser';
 import * as vegaLite from 'vega-lite';
 import {Config} from 'vega-lite/src/config';
 import {TopLevelSpec} from 'vega-lite/src/spec';
@@ -153,6 +155,10 @@ function parseVega(
 
   try {
     const spec = JSON.parse(action.spec);
+    const parsed = schemaParser(spec.$schema);
+
+    if (!satisfies(version, `^${parsed.version.slice(1)}`))
+      currLogger.warn(`The input spec uses Vega ${parsed.version}, but the current version of Vega is v${version}.`);
 
     validateVega(spec, currLogger);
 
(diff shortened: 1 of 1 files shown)


src/reducers/index.ts
```diff
@@ -215,6 +221,14 @@ function parseVegaLite(
       config,
       logger: currLogger
     };
+
+    const parsed = schemaParser(vegaLiteSpec.$schema);
+
+    if (!satisfies(vegaLite.version, `^${parsed.version.slice(1)}`))
+      currLogger.warn(
+        `The input spec uses Vega ${parsed.version}, but the current version of Vega is v${vegaLite.version}.`
```
domoritz:
```suggestion
        `The input spec uses Vega-Lite ${parsed.version}, but the current version of Vega-Lite is v${vegaLite.version}.`
```

adrijshikhar: done.

domoritz:
@siddhant1 Can you take a look at this?

siddhant1:
Screenshot 2020-02-19 at 3 17 00 PM

Should we check for the entire version or just the major release, if its the latter, the pr looks goos

adrijshikhar:
> Screenshot 2020-02-19 at 3 17 00 PM
> 
> Should we check for the entire version or just the major release, if its the latter, the pr looks good
> 
> Since we only host specific versions, there should be a warning when the URL is not resolvable, the editor does give a warning

Do you mean that there should be a warning when user uses v4.1 version instead of v4 as spec?

domoritz:
We should check for compatibility. Can you list a few examples of versions of the spec and the library that you tested (and whether they should be compatible or not)?

adrijshikhar:
i did not check the specs list. i simply checked the vega version with the input spec.

domoritz:
I asked you to list the cases you tried out to confirm that this feature is implemented correctly.

adrijshikhar:
now i added minor version check also. If a user enters `5.0` as spec, no warning is thrown as in the editor but if he enters `5.1` warning is thrown.

domoritz:
I don't understand exactly what the behavior is now. Can you please provide a list of cases so I can evaluate whether this is correct?

domoritz:
@adrijshikhar Do you plan to finish this pull request?

adrijshikhar:
My exams are going right now, they will be over on 8th march. As soon as they get over, i will start working on this.

domoritz:
Good luck with your exams!

domoritz:
Can you finish this @adrijshikhar?

adrijshikhar:
Yes i will get right on it.

adrijshikhar:
i am closing this PR and i will be opening another one

PR #619, opened 2020-02-19
Load gist specs using raw urls

### LINKED ISSUE #618: Long gist files are not fully read
I encountered an issue when loading very long specs from a Gist. (There already was some discussion on Slack: [Link](https://vega-js.slack.com/archives/C3N418BQD/p1581452305004600))

You can already see the same problem if you open the Gist on Github: [Link to Gist](https://gist.github.com/keckelt/74d3d57e1bbcd6182fe5a5cc555ab83e)
It contains to files:

* `failing.vl.json` which is truncated unless you open the Raw file:
https://files.slack.com/files-pri/T0NGA7XHN-FTKG7EYCS/image.png
*  And the shorter `tpm.vl.json` which loads fine 

The spec is read from the gist in this line: https://github.com/vega/editor/blob/master/src/components/header/gist-modal/renderer.tsx#L220

As you can see, it reads the `content` property from the json. But this content is truncated just like it is displayed truncated on Github.

The incomplete spec results in a _JSON is syntactically incorrect_ error in the UI and a SyntaxError in the console:
```
SyntaxError: "JSON.parse: end of data after property value in object at line 17561 column 24 of the JSON data"
    setGist app.tsx:123
    a tslib.es6.js:70
app.tsx:135:14

```

To fix the issue, the spec should be read raw, from the `raw_url` that is also part of the same JSON:
image
URL: https://api.github.com/gists/74d3d57e1bbcd6182fe5a5cc555ab83e/257ad5bb1148633b2d173acec4f0ffcff6bc6280

I will propose a PR :)

keckelt:
Fixes #618 

Loads GitHub Gist files from their raw url instead of reading the content from the metadata JSON, e.g.: https://api.github.com/gists/74d3d57e1bbcd6182fe5a5cc555ab83e/257ad5bb1148633b2d173acec4f0ffcff6bc6280

Large files are truncated in this JSON, therefore the specs should be read from the raw_url.


src/components/app.tsx
```diff
@@ -113,14 +113,16 @@ class App extends React.PureComponent<Props & {match: any; location: any; showEx
...
         `https://api.github.com/gists/${parameter.id}${
           parameter.revision !== undefined ? `/${parameter.revision}` : ''
         }`
       );
-      const data = await response.json();
-      const content = data.files[parameter.filename].content;
-      const contentObj = JSON.parse(content);
+      const gistData = await gistResponse.json();
+      const contentResponse = await fetch(gistData.files[parameter.filename].raw_url); // fetch from raw_url to handle large files
+      const content = await contentResponse.text(); // get as text
```
domoritz: can you just read `.json`?

src/components/app.tsx
```diff
@@ -113,14 +113,16 @@ class App extends React.PureComponent<Props & {match: any; location: any; showEx
...
+      const gistResponse = await fetch(
         `https://api.github.com/gists/${parameter.id}${
           parameter.revision !== undefined ? `/${parameter.revision}` : ''
         }`
       );
-      const data = await response.json();
-      const content = data.files[parameter.filename].content;
-      const contentObj = JSON.parse(content);
+      const gistData = await gistResponse.json();
+      const contentResponse = await fetch(gistData.files[parameter.filename].raw_url); // fetch from raw_url to handle large files
```
domoritz: Please remove comments that explain what the code does (and not why it does it)

src/components/header/gist-modal/renderer.tsx
```diff
@@ -215,17 +215,20 @@ class GistModal extends React.PureComponent<Props, State> {
...
-            } else {
-              this.props.history.push(`/gist/${gistId}/${revision}/${filename}`);
-            }
-            closePortal();
           }
+
+          return fetch(json.files[this.state.gist.filename].raw_url); // fetch from raw_url to handle large files
+        }
+      })
+      .then(async res => {
```
domoritz: Let's use await instead of then.

src/components/app.tsx
```diff
@@ -113,14 +113,16 @@ class App extends React.PureComponent<Props & {match: any; location: any; showEx
...
-      const response = await fetch(
+      const gistResponse = await fetch(
         `https://api.github.com/gists/${parameter.id}${
           parameter.revision !== undefined ? `/${parameter.revision}` : ''
         }`
       );
-      const data = await response.json();
-      const content = data.files[parameter.filename].content;
-      const contentObj = JSON.parse(content);
+      const gistData = await gistResponse.json();
```
domoritz: Does this mean we load small files multiple times? Can we avoid that?

keckelt: All gists are loaded multiple times if you use the dialogue
1. in `src/components/header/gist-modal/renderer.tsx` to check if it is valid
2. in `src/components/app.tsx` to actually use it

In `src/components/header/gist-modal/renderer.tsx` you could fall back to read from the raw url if the json is invalid, and thus potentially truncated - but that didn't feel quite right.

In `src/components/app.tsx` it should be fine to read directly from the raw url.

src/components/header/gist-modal/renderer.tsx
keckelt: Sure, i just wanted to stick to the code style there :)

src/components/app.tsx
keckelt: Then I have to stringify the json below

src/components/app.tsx
domoritz: Makes sense. Thanks for the explanation.

src/components/header/gist-modal/renderer.tsx
domoritz: Thanks for looking out for the code style. I like to use every opportunity for refactoring.

src/components/app.tsx
domoritz: Ahh, missed that. Thanks

domoritz:
@keckelt are you working on finishing these changes?

src/components/app.tsx
keckelt: I encountered one problem in `src/components/app.tsx`:
To directly read from the raw url I need the username, e.g.:
https://gist.githubusercontent.com/keckelt/83f0fdce4fcd5b758ba16e8570f76d92/raw/d1c9afb38a4129df7a6ef2963667c45032ba020b/Bin_Transform.vl.json

But the parameters there are taken from the URL (`https://vega.github.io/editor/#/gist/83f0fdce4fcd5b758ba16e8570f76d92/Bin_Transform.vl.json`) so it is not available there.

src/components/header/gist-modal/renderer.tsx
keckelt: Changed that bit

src/components/header/gist-modal/renderer.tsx
```diff
@@ -166,7 +166,7 @@ class GistModal extends React.PureComponent<Props, State> {
         });
         return res.json();
       })
-      .then(json => {
+      .then(async json => {
```
domoritz: Use `await` instead of `then`

keckelt: I've refactored the whole function to use async/await.
The code diff is less/displayed better in VS Code

src/components/app.tsx
domoritz: Do you mean we could refactor to directly load the file if we knew the username? If so, can you file an issue about this?

src/components/header/gist-modal/renderer.tsx
```diff
@@ -132,109 +132,106 @@ class GistModal extends React.PureComponent<Props, State> {
...
           this.setState({
             gistLoadClicked: false,
-            syntaxError: true
+            invalidFilename: true
           });
+          return Promise.reject('Invalid file name');
         }
-      });
+
+        const rawResponse = await fetch(gistSummary.files[this.state.gist.filename].raw_url); // fetch from raw_url to handle large files
```
domoritz: What is this used for?

src/components/header/gist-modal/renderer.tsx
```diff
@@ -132,109 +132,106 @@ class GistModal extends React.PureComponent<Props, State> {
...
-      .catch(error => {
-        if (error instanceof SyntaxError) {
+      } else {
+        if (gistSummary.files[this.state.gist.filename] === undefined) {
           this.setState({
             gistLoadClicked: false,
-            syntaxError: true
+            invalidFilename: true
           });
+          return Promise.reject('Invalid file name');
```
domoritz: Why are we rejecting a promise instead of raising an error?

domoritz: We are in an async function so raising an error should do the same.

src/components/header/gist-modal/renderer.tsx
```diff
@@ -132,109 +132,106 @@ class GistModal extends React.PureComponent<Props, State> {
...
         }
-      });
+
+        const rawResponse = await fetch(gistSummary.files[this.state.gist.filename].raw_url); // fetch from raw_url to handle large files
+
+        const {revision, filename} = this.state.gist;
+        await rawResponse.json(); // check if the loaded file is a JSON
```
domoritz: move this up two lines

src/components/header/gist-modal/renderer.tsx
keckelt: right, i forgot changing those as well.

domoritz: Let's change them before merging.

src/components/header/gist-modal/renderer.tsx
domoritz: So it's with the fetch statement since the variable doesn't get used anymore.

src/components/header/gist-modal/renderer.tsx
domoritz: thanks!

src/components/app.tsx
keckelt: yes.
I assume you would need a second gist route like `https://vega.github.io/editor/#/gist/<username>/<gist-hash>/<filename>`
e.g.:
`https://vega.github.io/editor/#/gist/keckelt/83f0fdce4fcd5b758ba16e8570f76d92/Hist with null.vl.json`

domoritz: Ahh, no, we don't want that probably. I was thinking we could use the user name of the logged in user but that's only true for some gists.

src/components/header/gist-modal/renderer.tsx
keckelt: This one was passed as callback but that didn't work properly (blank page was displayed) so I moved it out of the function call:
https://github.com/vega/editor/pull/619/commits/46754d39373990f4606052088c954aef9e1459ca#diff-d5eed843722bc0e9999038c61414bb4dL183-L185

domoritz:
Thank you @keckelt

Issue #611, opened 2020-02-18
should open to new tab


shakti97:
clicking on vega view open the link on the same tab, i think it should open the link in another tab so that one should not have to leave vega app 
![Screenshot from 2020-02-18 10-08-10](https://user-images.githubusercontent.com/30585570/74704573-e1e7d500-5236-11ea-9e4e-ea35f6bb0fcd.png)

There are other links also like hover link in this image
![Screenshot from 2020-02-18 10-11-47](https://user-images.githubusercontent.com/30585570/74704688-528ef180-5237-11ea-8418-95958da5e6c5.png)

shakti97:
if possible would like to work on it @domoritz

siddhant1:
Thanks for reporting @shakti97 , you can go on and fix this

domoritz:
Hmm, you can CMD+click on any link to open them in a new tab but as far as I know there is no way to do the reverse. I’m not yet convinced we want this change.

siddhant1:
The other links do open in new tab , we should keep it consistent though

domoritz:
Ahh, yes, we should. Let’s open them all in new tabs or the same tab then. I can go either way.

ashu8912:
@domoritz  I would recommend opening every link in the same tab 
Here are the reasons why
1.If a user prefers that links open in a new tab, they can open the link in a new tab using a right-click or by continuing to press down on the link on a mobile device. This option is always available to users no matter how a link is set to open.
However, it doesn’t work in reverse.As you also mentioned.
2.The only time it is recommended that you open a link in a new tab is when opening in the same screen would interrupt a process (which in our case isn't the case or is it????)
3.From the usability point of view the decision to enforce opening links in new windows violates one of the fundamental principles of the user interface design: users should always be in control of the interface they are interacting with.
The above are the reasons why i think they should open in the same tab If any other suggestion or other outcomes of opening in same tab are there they are most welcomed for discussions.
Also as @siddhant1 mentioned there should be consistency all around(meaning it should not be as if one link opens in the same tab and other open in other tabs coz this leads users to a surprise to links clicking).

shakti97:
When you open external links in the same tab, you create back-button fatigue for users. Every time the user goes to an external website they have to hit the back button to go back to your app. If they decide to click the links on the external page/website, they have to hit the back button even more times to get back to your site. This is a lot of unnecessary work for users. So I think it should be open in a new tab.

ashu8912:
[...]

PR #625, opened 2020-02-22
fix link target consistency

shakti97:
The fix for issue #611  has missed this instance


shakti97:
@domoritz please have a look and let me know if any changes is required:)

Issue #627, opened 2020-02-22
Typo issue in reset


shakti97:
![Screenshot from 2020-02-23 03-05-00](https://user-images.githubusercontent.com/30585570/75099618-a871e980-55e9-11ea-8e06-7855343d3577.png)

## stuck instead of struck

shakti97:
@domoritz , if possible can i work on this?

domoritz:
Thanks. Feel free to just send prs for small issues like this.

PR #628, opened 2020-02-22
fix typo issue on reset

shakti97:
It will fix issue #627


shakti97:
@domoritz have a look and let me know if any changes required:)

PR #629, opened 2020-02-24
Add VS Code format rule to remove spaces after opening and before closing non-empty braces

keckelt:
With the `"editor.formatOnSave": true` [in the settings.json](https://github.com/vega/editor/blob/master/.vscode/settings.json#L3) my VS Code introduced a lot of spaces inside the curly braced object definitions so I've added this setting to keep the currently present code format:
```json
"typescript.format.insertSpaceAfterOpeningAndBeforeClosingNonemptyBraces": false, 
```
> Defines space handling after opening and before closing non-empty braces. Requires using TypeScript 2.3.0 or newer in the workspace.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,7 @@
 {
   "typescript.tsdk": "node_modules/typescript/lib",
   "editor.formatOnSave": true,
+  "typescript.format.insertSpaceAfterOpeningAndBeforeClosingNonemptyBraces": false,
   "eslint.validate": [
     "javascript",
     "javascriptreact",


keckelt:
Didn't read the contributing guudelines careful enough. I'll try it later.

keckelt:
With prettier the format is not changed on save and without i get this warning when i try to format the code:
image

domoritz:
Can you install the prettier extension?

keckelt:
Yes, I did. 
With 

> the format is not changed

I meant it's formatted ass it should be after installing prettier.

domoritz:
Nice.

PR #658, opened 2020-04-06
Add restart option into docker compose yml file

y-vectorfield:
I added `restart "always"` option into docker-compose.yml. If we add this option, we don't need to restart the container when we reboot the machine.


y-vectorfield:
@domoritz
Hello, Dr. Moritz, 
Thank you very much merging my pull request!

PR #691, opened 2020-05-06
Functionality for creating and updating GitHub GIST

algomaster99:
A continuation of https://github.com/vega/editor/pull/489

These changes introduce:
- Create GIST directly from editor
- Update GIST directly from editor

### Creation of GIST
1. Go to `SHARE` modal in the header.
2. Fill out:
    1. Check "Private gist" if you want to make it a private gist otherwise it is created as a public GIST.
    2. Write the file name. (you may or may not add `.json` as the extension. eg. `graph` and `graph.json` are equivalent).
    3. The title is the description of the GIST which is optional.
3. Once you click `Create`, its text will change to `Creating...` and in around 5 seconds your GIST will be created (depending upon the internet connection of course).

### Updating GIST
1. Select the GIST according by its description in the dropdown list which is to the left of the checkbox.
2. Once you select a particular gist, it will fetch the details of only the first file in the gist. (won't replace the `editorString`).
3. You may edit the `title`, `filename` but you cannot edit its privacy since the [API](https://developer.github.com/v3/gists/#update-a-gist) doesn't allow it.
4. Once you click `UPDATE`, it will replace the gist content with `editorString` you have currently.

### Demonstration
vega

#### NOTE
1. The UI/UX is not finalized. @domoritz , I would request you to give suggestions where we can put these two functionalities.
2. Like I mentioned that only the first file of the GIST can be edited, do we want to extend this support so that any file inside a particular gist can be edited? For that, I think we can add the section for creating and updating GIST below the `LOAD GIST` in this modal.
Screenshot from 2020-05-06 15-09-59
From this, we can benefit from selecting any file which is shown on the left side of the modal. 
3. I also had to resort to a `workaround` when I implement the API for updating. From `vega-editor` to `vega-editor-backend`, I am sending a `POST` request for sending a `PATCH` request for updating the GIST. I was trying to make the former also as `PATCH` request, but I kept getting `CORS` error.


domoritz:
- [ ] I pushed some improvements. The backend returns an empty response on error, which I think is confusing. Let's fix that. 

- [ ] And then one other issue is that the title is optional. Let's have a default placeholder that is `Vega Spec from May 6, 2020` or `Vega-Lite Spec from May 6, 2020`.

Then we are good to merge.

algomaster99:
> 2\. Like I mentioned that only the first file of the GIST can be edited, do we want to extend this support so that any file inside a particular gist can be edited? For that, I think we can add the section for creating and updating GIST below the `LOAD GIST` in this modal.

@domoritz are you okay with this though?

domoritz:
As a user, I can customize the file name, right? Maybe one simple solution would be to use a nested dropdown where we show all possible files for the gists that are loaded. Something like

Screen Shot 2020-05-06 at 14 49 57

Yes, I agree that supporting the selector UI you already built would be better, though. Can you make it into a reusable component that we can use for loading and saving gists?

mattijn:
I'm not sure if I'm too late at the table, but I've a small UI comment. When I'm working with vega(-lite) specs (and any other file on PC), I distinct them based on filenames and not on file descriptions. So once loaded in the vega editor, my gist specs looks like this:
image
Almost none of them have a description, but all have a filename. Maybe I'm opposing design criteria, but would it be possible to swap this? So in bold, the filename and as child node the spec description?

algomaster99:
@domoritz I separated the widget for gist selection as discussed in the call today.

algomaster99:
@domoritz 

- I have returned the URL of the created/updated gist in response. I think we should render it somewhere. Let me know where.
- I have 0 design sense so please feel free to modify the styles of the gist when it is highlighted :sweat_smile: 

> * And then one other issue is that the title is optional. Let's have a default placeholder that is `Vega Spec from May 6, 2020` or `Vega-Lite Spec from May 6, 2020`.

Since we don't have the drop-down list now, is it necessary?

domoritz:
* [ ] fix the warning from https://lgtm.com/projects/g/vega/editor/rev/pr-d7818dd117813dbbd0e828d8e96fc5c7e15e8783
* [ ] Don't fill in the selected file into the "new gist" side when I update a gist
* [ ] Add default title for new gist. 
* [ ] For new gists, move the tile input above the file name. 

Once that's done, I can improve the style myself.

PR #697, opened 2020-05-08
Add field for creating new files in Gist through update

algomaster99:


src/components/header/share-modal/renderer.tsx
```diff
@@ -303,6 +301,20 @@ class ShareModal extends React.PureComponent<Props, State> {
...
             <GistSelectWidget selectGist={this.selectGist.bind(this)} />
+            {this.props.isAuthenticated && (
+              <div className="share-input-container">
+                <label>
+                  File name: (optional)
+                  <input
+                    value={this.state.gistFileNameNew}
+                    name="gistFileNameNew"
+                    onChange={this.textChange.bind(this)}
+                    placeholder="Fill this only if you want to create a new file in the selected gist"
```
domoritz, line 312: That's not what placeholders are for. Let's make this a text description below. The placeholder should be the original file name (if any).

PR #707, opened 2020-05-26
Add base url support for render service

### LINKED ISSUE #686: Correctly support base urls in the PDF generation
Right now, charts that use relative URLs don't render because the data is not loaded. 

See for example https://vega.github.io/editor/#/examples/vega-lite/stacked_bar_weather.

hungngo97:
Fix #686 to support base url on render service


src/components/header/export-modal/renderer.tsx
```diff
@@ -46,8 +46,13 @@ class ExportModal extends React.PureComponent<Props, State> {
     this.setState({loadingPDF: true});
 
     const content = this.props.mode === Mode.Vega ? this.props.vegaSpec : this.props.vegaLiteSpec;
+    console.log('Base url', this.props.baseURL);
```
domoritz, line 49: remove

src/components/header/export-modal/renderer.tsx
```diff
@@ -46,8 +46,13 @@ class ExportModal extends React.PureComponent<Props, State> {
     this.setState({loadingPDF: true});
 
     const content = this.props.mode === Mode.Vega ? this.props.vegaSpec : this.props.vegaLiteSpec;
```
domoritz, line 48: rename to `spec`

src/components/header/export-modal/renderer.tsx
domoritz, line 49:
```suggestion
```

PR #744, opened 2020-07-13
feat: attach current VL/VG specs to VEGA_DEBUG

arvind:


src/components/renderer/renderer.tsx
```diff
@@ -166,7 +166,11 @@ class Editor extends React.PureComponent<Props, State> {
 
     (view as any).logger(dispatchingLogger);
 
-    (window as any).VEGA_DEBUG.view = view;
+    const debug = (window as any).VEGA_DEBUG;
+    debug.vegaSpec = vegaSpec;
+    debug.vegaLiteSpec = vegaLiteSpec;
```
domoritz, line 171: We should probably unset this when we are looking at a Vega spec. Right now, when you are looking at a Vega chart, you still get the Vega-Lite spec.

arvind: Good call. Force-pushed a change.

Issue #772, opened 2020-08-18
Download JSON should use compact json stringify


domoritz:
https://github.com/vega/editor/blob/378bfea22e288aea25257271fe97df4a9f46d00f/src/components/header/export-modal/renderer.tsx#L105

domoritz:
Fixed in https://github.com/vega/editor/pull/820

PR #820, opened 2020-10-04
Compact json download

ashu8912:


ashu8912:
fixes #772

src/components/header/export-modal/renderer.tsx
```diff
@@ -102,7 +102,7 @@ class ExportModal extends React.PureComponent<Props, State> {
       content.config = mergeConfig({}, this.props.config, content.config);
     }
 
-    const blob = new Blob([JSON.stringify(content, null, 2)], {
+    const blob = new Blob([JSON.stringify(content, null, 0)], {
```
domoritz, line 105: Please use json-stringify-pretty-compact instead.

domoritz (changes requested):

src/components/header/export-modal/renderer.tsx
ashu8912: aah nice we want best of both the worlds, i will add the fix

ashu8912: also @domoritz do we want that 0 indentation or the previous default 2 indentation

domoritz: We don't need any arguments anymore. Look at other uses of json-stringify-pretty-compact.

ashu8912:
@domoritz fixed and tested as well

Issue #885, opened 2020-12-21
Error thrown when using tooltips and dynamic data 


GrzegorzKazana:
Hi 👋 

I came across an issue when using tooltips and dynamic data - it seemed to me it might be related to how the tooltips are handled in the editor, so I am opening the issue here - feel free to move it is not the right place.

### Expected behavior

When having tooltips enabled and using dynamic data - by defining `url` by a signal value (as described [here](https://vega.github.io/vega/docs/data/) - Dynamic Data Loading section), no errors should be thrown.

### Actual behavior

When enabling tooltips and using dynamic data, editor crashes with the following error
```
Cannot use 'in' operator to search for 'message' in Dataflow already running. Use runAsync() to chain invocations.
```

### Steps to reproduce

1. Enable tooltips: `Settings -> Tooltips` checkbox
2. Use specification with dynamic data, i.e.
```js
{
  "signals": [
    {
      "name": "testtest",
      "value": "https://jsonplaceholder.typicode.com/posts"
    }
  ],
  "data": [
    {
      "name": "baseData",
      "url": {"signal": "testtest"}
    }
  ]
}
```
3. Run the view (if not run automatically)
4. (An error is raised)
5. Disable the tooltips, and rerun the view
6. everything works fine

##### [Minimal reproduction (link to editor)](https://vega.github.io/editor/#/url/vega/N4Igzglg5gdghgGzCAXAbVPAtgU1SAFxzAKJJABoQA3RAVzxRAAtSAHMFAei4CswA9jDYI4AYxzMBCACY4ATgDoCATzYQxAuYs1YubASWQBfALpUZcAnFQYQ2RiABGcMDgAiVm1TryEqUEhYRHwyUmICEGMzYyA)

```
Versions
Vega: 5.17.0
Vega-Lite: 4.17.0
Vega-Tooltip: 0.24.2
Editor: 0.91.2

Chrome 87.0
Firefox 83.0
```

domoritz:
Is this a problem with the Vega editor or Vega or Vega Tooltip?

GrzegorzKazana:
It is difficult for me to say, thought it might be related how the tooltip is called from the editor app (?)
Feel free to correct me 🤷

domoritz:
Could you try to reproduce the example without the editor? You can start with https://bl.ocks.org/domoritz/cd636b15fa0e187b51b73fc60b4d3014.

GrzegorzKazana:
[...]

PR #887, opened 2020-12-21
fix: tooltip initalization by awaiting running the view

GrzegorzKazana:
Addresses https://github.com/vega/editor/issues/885.

Changes:
- initialize tooltips after the view is run asynchronously


domoritz (changes requested):
Please change to async ... await and I will merge. 

Screen Shot 2020-12-21 at 16 13 29

domoritz:
Thank you @GrzegorzKazana

PR #1023, opened 2021-05-27
Add Dataflow Viewer

saulshanabrook:
This PR adds a dataflow viewer, to assist in debugging Vega visualizations.

It currently uses @jheer's [`vega2dot`](https://observablehq.com/@vega/vega2dot) which is used in the ["How Vega Works"](https://observablehq.com/@vega/how-vega-works) guide.

Screen Shot 2021-05-26 at 9 36 23 PM

Instead, we could also base this off of @chengluyu's [`vega-inspector`](https://github.com/chengluyu/vega-inspector). I haven't looked at the code yet. @domoritz also wrote a dataflow visualizer (https://github.com/vega/dataflow-vis) a while ago

TODO:

- [ ] Try out using web worker, to move Graphviz compilation off the main thread
- [ ] Possibly add ability to step through timestamps, as shown in the How Vega Works guide
- [ ] ??? other UX improvements to make the graph more usable ???
- [ ] Maybe make graph interactive, instead of rendering to SVG use a JS graph viewer. Could use [Eclipse Layout Kernel](https://github.com/kieler/elkjs) like [ipyelk](https://github.com/jupyrdf/ipyelk#screenshots) does.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "d3-geo-projection": "^3.0.0",
     "d3-scale": "^3.3.0",
     "file-loader": "^6.2.0",
+    "graphviz-react": "1.2.0",
     "history": "^5.0.0",
     "json-stringify-pretty-compact": "^3.0.0",
     "lz-string": "^1.4.4",
--- a/src/components/dataflow-viewer/index.css
+++ b/src/components/dataflow-viewer/index.css
@@ -1,0 +1,5 @@
+.dataflow-pane {
+  overflow: auto;
+  width: 100%;
+  flex: 1;
+}
--- a/src/components/dataflow-viewer/index.tsx
+++ b/src/components/dataflow-viewer/index.tsx
@@ -1,0 +1,12 @@
+import {connect} from 'react-redux';
+import {State} from '../../constants/default-state';
+import Renderer from './renderer';
+
+export function mapStateToProps(state: State) {
+  return {
+    editorRef: state.editorRef,
+    view: state.view,
+  };
+}
+
+export default connect(mapStateToProps)(Renderer);
--- a/src/components/dataflow-viewer/renderer.tsx
+++ b/src/components/dataflow-viewer/renderer.tsx
@@ -1,0 +1,55 @@
+import {Graphviz} from 'graphviz-react';
+import * as React from 'react';
+import {useEffect, useRef, useState} from 'react';
+import {mapStateToProps} from '.';
+import {view2dot} from '../../utils/view2dot';
+import './index.css';
+
+type StoreProps = ReturnType<typeof mapStateToProps>;
+
+export default function DataflowViewer({view}: StoreProps) {
+  const componentRef = useRef();
+
+  const dot = view2dot(view, 0);
+  // TODO: Use webworker for graphviz
+  // https://github.com/DomParfitt/graphviz-react/issues/37
+
+  // Can't set to parent width, so manually get parent container and use this to set
+  // https://github.com/DomParfitt/graphviz-react/issues/11
+  const {width, height} = useContainerDimensions(componentRef);
+  return (
+    <div ref={componentRef} className="dataflow-pane">
+      <Graphviz dot={dot} options={{fit: true, zoom: true, width, height}} />
+    </div>
+  );
+}
+
+/**
+ * From https://stackoverflow.com/a/60978633/907060
+ */
+export const useContainerDimensions = (myRef) => {
+  const getDimensions = () => ({
+    width: myRef.current.offsetWidth,
+    height: myRef.current.offsetHeight,
+  });
+
+  const [dimensions, setDimensions] = useState({width: 0, height: 0});
+
+  useEffect(() => {
+    const handleResize = () => {
+      setDimensions(getDimensions());
+    };
+
+    if (myRef.current) {
+      setDimensions(getDimensions());
+    }
+
+    window.addEventListener('resize', handleResize);
+
+    return () => {
+      window.removeEventListener('resize', handleResize);
+    };
+  }, [myRef]);
+
+  return dimensions;
+};
--- a/src/components/viz-pane/debug-pane-header/index.tsx
+++ b/src/components/viz-pane/debug-pane-header/index.tsx
@@ -71,6 +71,20 @@ class DebugPaneHeader extends React.PureComponent<Props> {
               Signal Viewer
             </li>
           )}
+          {error === null && (
+            <li
+              className={navItem === NAVBAR.DataflowViewer ? 'active-tab' : undefined}
+              onClick={(e) => {
+                if (debugPane) {
+                  e.stopPropagation();
+                }
+                showLogs(false);
+                toggleNavbar(NAVBAR.DataflowViewer);
+              }}
+            >
+              Dataflow Viewer
+            </li>
+          )}
         </ul>
         {debugPane ? <ChevronDown /> : <ChevronUp />}
       </div>
--- a/src/components/viz-pane/renderer.tsx
+++ b/src/components/viz-pane/renderer.tsx
@@ -1,17 +1,18 @@
 import * as React from 'react';
 import SplitPane from 'react-split-pane';
+import {version as VG_VERSION} from 'vega';
+import {version as VL_VERSION} from 'vega-lite';
+import {version as TOOLTIP_VERSION} from 'vega-tooltip';
 import {mapDispatchToProps, mapStateToProps} from '.';
 import {EDITOR_FOCUS, LAYOUT, NAVBAR, WORD_SEPARATORS} from '../../constants';
 import DataViewer from '../data-viewer';
+import DataflowViewer from '../dataflow-viewer';
 import ErrorBoundary from '../error-boundary';
 import ErrorPane from '../error-pane';
 import Renderer from '../renderer';
 import SignalViewer from '../signal-viewer';
 import DebugPaneHeader from './debug-pane-header';
 import './index.css';
-import {version as VG_VERSION} from 'vega';
-import {version as VL_VERSION} from 'vega-lite';
-import {version as TOOLTIP_VERSION} from 'vega-tooltip';
 const pjson = require('../../../package.json');
 
 const defaultState = {
@@ -98,6 +99,8 @@ export default class VizPane extends React.PureComponent<Props, State> {
           return <DataViewer onClickHandler={(header) => this.onClickHandler(header)} />;
         case NAVBAR.SignalViewer:
           return <SignalViewer onClickHandler={(header) => this.onClickHandler(header)} />;
+        case NAVBAR.DataflowViewer:
+          return <DataflowViewer />;
         default:
           return null;
       }
--- a/src/constants/consts.ts
+++ b/src/constants/consts.ts
@@ -47,6 +47,7 @@ export const NAVBAR = {
   DataViewer: 'DataViewer',
   Logs: 'Logs',
   SignalViewer: 'SignalViewer',
+  DataflowViewer: 'DataflowViewer',
 } as const;
 
 export const SIDEPANE = {
--- a/src/utils/view2dot.ts
+++ b/src/utils/view2dot.ts
@@ -1,0 +1,194 @@
+// Copy of https://observablehq.com/@vega/vega2dot
+
+/**
+ * Generates a dot-formatted file for Graphviz to visualize the dataflow
+ * graph for a Vega View instance.
+ * Note: does not (yet) support nested scopes / subflows.
+ * @param {View} view A Vega View instance.
+ * @param {number} [stamp] Optional pulse timestamp. If provided, dataflow
+ *   nodes and edges that have not been evaluated since the timestamp will
+ *   be deemphasized.
+ * @returns {string} The generated dot file.
+ */
+export function view2dot(view, stamp) {
+  const rt = view._runtime,
+    ops = rt.nodes,
+    keys = Object.keys(ops);
+
+  const signals = Object.keys(rt.signals).reduce((lut, name) => {
+    lut[rt.signals[name].id] = name;
+    return lut;
+  }, {});
+
+  const scales = Object.keys(rt.scales).reduce((lut, name) => {
+    lut[rt.scales[name].id] = name;
+    return lut;
+  }, {});
+
+  const data = Object.keys(rt.data).reduce((lut, name) => {
+    const sets = rt.data[name];
+    if (sets.input) lut[sets.input.id] = name;
+    if (sets.output) lut[sets.output.id] = name;
+    return lut;
+  }, {});
+
+  // build node objects
+  const nodes = keys.map((key) => {
+    const op = ops[key];
+    const node: any = {
+      id: op.id,
+      type: op.constructor.name.toLowerCase(),
+      stamp: op.stamp,
+      value: op,
+    };
+    if (markTypes[getType(node.type)]) node.isMark = true;
+    if (signals[op.id]) node.signal = signals[op.id];
+    if (scales[op.id]) node.scale = scales[op.id];
+    if (data[op.id]) node.data = data[op.id];
+    if (rt.root === op) node.root = true;
+    return node;
+  });
+
+  const ids = nodes.reduce((lut, node) => {
+    lut[node.id] = node;
+    return lut;
+  }, {});
+
+  // build edge objects
+  const edges = [];
+  keys.forEach((key) => {
+    const op = ops[key];
+    if (op._targets)
+      op._targets.forEach((t) => {
+        if (!ids[t.id]) return;
+        edges.push({
+          nodes: [op.id, t.id],
+          param: t.source === op ? 'pulse' : argop(t, op),
+        });
+        if (t.source === op && ids[op.id].isMark) {
+          const node = ids[t.id];
+          if (getType(node.type) === 'collect') {
+            // annotate post-datajoin collect operators as mark-processing
+            node.isMark = true;
+          }
+        }
+      });
+  });
+
+  return `digraph {
+    rankdir = LR;
+    node [style=filled];
+    ${nodes
+      .map((node) => {
+        return (
+          node.id +
+          ' [label="' +
+          nodeLabel(node) +
+          '"]' +
+          ' [color="' +
+          nodeColor(node, stamp) +
+          '"]' +
+          ' [fillcolor="' +
+          nodeFillColor(node, stamp) +
+          '"]' +
+          ' [fontcolor="' +
+          nodeFontColor(node, stamp) +
+          '"]'
+        );
+      })
+      .join(';\n  ')};
+    ${edges
+      .map((e) => {
+        return (
+          e.nodes.join(' -> ') +
+          ' [label="' +
+          (e.param === 'pulse' ? '' : e.param) +
+          '"]' +
+          ' [color="' +
+          edgeColor(e, ids, stamp) +
+          '"]' +
+          ' [fontcolor="' +
+          edgeLabelColor(e, ids, stamp) +
+          '"]' +
+          ' [weight="' +
+          edgeWeight(e, ids) +
+          '"]'
+        );
+      })
+      .join(';\n  ')};
+  }`;
+}
+
+function nodeLabel(node) {
+  return node.signal
+    ? node.signal
+    : node.scale
+    ? node.scale
+    : node.root
+    ? 'root'
+    : getType(node.type) === 'collect' && node.data
+    ? node.data
+    : getType(node.type);
+}
+
+function nodeFillColor(node, stamp) {
+  return stamp && node.value.stamp < stamp
+    ? '#ffffff'
+    : node.signal
+    ? '#dddddd'
+    : node.scale
+    ? '#ccffcc'
+    : node.data
+    ? '#ffcccc'
+    : getType(node.type) === 'axisticks' || getType(node.type) === 'legendentries'
+    ? '#ffcccc'
+    : node.isMark || node.root
+    ? '#ccccff'
+    : '#ffffff';
+}
+
+function nodeColor(node, stamp) {
+  return stamp && node.value.stamp < stamp ? '#dddddd' : '#000000';
+}
+
+function nodeFontColor(node, stamp) {
+  return stamp && node.value.stamp < stamp ? '#cccccc' : '#000000';
+}
+
+function edgeColor(edge, nodes, stamp) {
+  const n = edge.nodes;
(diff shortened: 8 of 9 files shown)


domoritz:
Awesome. I think moving graphviz to a web worker, adding types, and moving towards the code standard we have in the editor (I can help with that) are the next steps and then we can iterate on it.

For rendering, we could also get just the locations from graphviz but render the nodes ourselves. There are also https://github.com/dagrejs/dagre and https://github.com/dagrejs/graphlib for doing layouts.

saulshanabrook:
I have also used [cytoscape JS](https://js.cytoscape.org/) in the past to do frontend rendering, which is nice in that it has adaptors for a bunch of layout engines (including dagre, ELK, and graphviz), so you can play with which one looks the most readable for your data.

The one thing I don't love about it is that it doesn't support "port" functionality (https://github.com/cytoscape/cytoscape.js-elk/issues/1), which is supported in ELK and can be handy!

chengluyu:
Brilliant!

_[How Vega Works](https://observablehq.com/@vega/how-vega-works)_ didn’t handle some cases. @z-jack extended the code (see [these commits](https://github.com/chengluyu/vega-inspector/commits?author=z-jack)) but I’m not sure if he considered all possible cases.

Me and @z-jack had discussed about a debug mode, in which people could pause the execution of their Vega programs and trace data. I think it is akin to your “step through timestamps” We didn’t finish because we were hacking Vega externally and it requires a lot of efforts. 😅

I tried Dagre (also suggested by @domoritz) in one of my other projects. It’s easier to use ([example](https://github.com/dagrejs/dagre/wiki#an-example-layout)) and has a straightforward API. In that project, I passed the layout information to [react-flow-renderer](https://github.com/wbkd/react-flow), which supports customizable and interactive graph nodes.

domoritz:
Thank you for your feedback @chengluyu. We actually discussed your viewer and were wondering about the differences. Thanks for the pointers. This is very useful to know.

saulshanabrook:
Thanks @chengluyu for the code links! I might try out the code you and @z-jack wrote to see how that looks.

I noticed that @z-jack changed the analysis to use the parsed runtime from `vega.parse` instead of looking directly at the materialized vega view (as jeff had done  originally) in chengluyu/vega-inspector@f66463f. 

That makes sense to me, but was curious why you chose to make the change or why Jeff had originally chosen to use the view!

saulshanabrook:
I was also thinking it might be useful to try adding the compiled dataflow runtime spec as a JSON pane on the bottom, like you can see the compiled Vega, when you are editing Vega Lite?

domoritz:
We already have VEGA_DEBUG in the browser console but I agree that a JSON view could be useful as well.

saulshanabrook:
Yeah I was using `VEGA_DEBUG.vega.parse(VEGA_DEBUG.vegaSpec)` but I added the dataflow to the viewer, at least temporarily to help me debug things!

For now, I added it to the top panel because it was easier than adding to the bottom.

domoritz:
You can access the view with `VEGA_DEBUG` as well, which gives you the runtime.

saulshanabrook:
I pulled in @chengluyu  and @z-jack's code to create the graphviz, which adds some more details.



I pulled in their repo, so they will be credited in the git log. 

```shell
git fetch git@github.com:chengluyu/vega-inspector.git master:inspector-master
git checkout inspector-master
git subtree split --prefix=src/ -b subtree-inspector-src
git checkout -
git subtree add --prefix=inspector_src subtree-inspector-src
git mv inspector_src/helpers/vega2dot.* src/utils/
git rm -r inspector_src/
```

src/utils/vega2dot.js
```diff
@@ -0,0 +1,1058 @@
+import {EventStream, Operator, Dataflow} from 'vega-dataflow';
```
domoritz, line 1: Let's write this in ts.

domoritz (changes requested):

src/utils/vega2dot.js
saulshanabrook: Sounds good. I just copied over what was previously in the `vega-inspector` for now.

saulshanabrook:
I have switched to use [Cytoscape JS](https://js.cytoscape.org/) with the [ELK layout engine](https://github.com/cytoscape/cytoscape.js-elk), to make it easier to navigate the graph.



I am still having some issues with the graph layout, but it seems more legible to me than before. I added the ability to click on a node and have it fit to its relatives.

saulshanabrook:
After some fixes (https://github.com/cytoscape/cytoscape.js-elk/pull/30, https://github.com/cytoscape/cytoscape.js-elk/pull/32) to the cytoscape ELK integration, the layouts seem accurate now to the edge dependencies.


https://user-images.githubusercontent.com/1186124/123444899-640fa880-d5a5-11eb-84fb-5284784712f3.mov

saulshanabrook:
@domoritz I had a question about the dataflow semantics of subflows. I see in the code that parses the subflow parameters that it treats the first operation in the subflow in a special manner, as well as the `parent` signal in the subflow:

https://github.com/vega/vega/blob/2d1462476d39889c31d23e5970296929ff2405bf/packages/vega-runtime/src/parameters.js#L128-L138

I have tried to understand that code, but I am still unclear what the `parent` signal is used for as well as what is special about the first operator.

As an example, here is a screenshot of part of a subfacet subflow from the crossfilter example. I have circled the first operator, a `sieve` operator with no relations, as well as the `parent` operator.

In this case, the parent signal is the parent of all the `mark` operators, which take in the `context`. One guess I had is that whenever the subflow operator itself receives a pulse, it is passed through into the node with the `parent` signal, which then triggers all the marks to update using their context to somehow retrieve the data?

---

I am contemplating adding some other debug view to the dataflow viewer to see records of all the pulses and what nodes they update. My thought was that this sort of debugger could help me diagnose these types of questions.

saulshanabrook:
If you had any thoughts on the interaction patterns in the graph, that would also be useful. Currently, if I include all the details, like I do in the above screenshot, it makes it a bit unwieldy. However, without any details, it's hard to get the right context. 

I already have on click overloaded to filter to a subset of the graph, but could change that to instead expand/collapse details. Or have a global toggle.

saulshanabrook:
I fixed the selection logic, so that we can select multiple nodes. It also now will select related compound nodes:

https://user-images.githubusercontent.com/1186124/125536378-20b05934-f5bb-49eb-ad82-68099d1d88d7.mov

domoritz:
I don't know the answer to https://github.com/vega/editor/pull/1023#issuecomment-879383219. @jheer @arvind can you help answer the questions?

> If you had any thoughts on the interaction patterns in the graph, that would also be useful. Currently, if I include all the details, like I do in the above screenshot, it makes it a bit unwieldy. However, without any details, it's hard to get the right context. 

One thing you might be able to do is to filter our nodes coming from axes/legends. It could reduce the complexity of the graphs significantly.

saulshanabrook:
FYI needs to wait on https://github.com/cytoscape/cytoscape.js-elk/pull/32 to be released. To try locally, I am just editing the file directly.

saulshanabrook:
I moved the details to a tooltip, which improves the layout, I think!

https://user-images.githubusercontent.com/1186124/127181568-5906300b-5143-4f51-b917-0582e18066ff.mov

I have a few other semantics questions that I want to try to answer and maybe incorporate the learning into the viewer, like:

1. Which signals are used to display the graph? I see the `padding` and `background` signals attached to isolated nodes often. To make the data dependence a little more explicit, we could have a root node that depends on all node signals that are used by the rendering engine 
Screen Shot 2021-07-27 at 11 39 22 AM

2. What is the purpose of the `sieve` operators with the data? They are often leaves in the graph. They set the data that is viewable in the data viewer? What is the purpose of registering data here? 
Screen Shot 2021-07-27 at 11 39 28 AM

domoritz:
@jheer and @arvind, could you answer these questions about the dataflow

saulshanabrook:
For running the layout in a webworker, I was talking to @ about trying to patch Cytoscape ELK to have it run with a webworker. I opened an upstream issue about it: https://github.com/cytoscape/cytoscape.js-elk/issues/34 

ELK JS looks like it has support for this: https://github.com/kieler/elkjs#browser, so it seems like if we instantiate that properly then the layout engine (https://github.com/cytoscape/cytoscape.js-elk/blob/ce1f11d8d9d472d92148f6ec101e69b40268c8b9/src/layout.js#L164-L165) can push computing the layout to the webworker. 

However, I haven't been able to get it to thread properly through all the build steps and webpack to server the webworker file and then find it. I did find this webpack documentation: https://webpack.js.org/guides/web-workers/

I have been using `npx patch-package cytoscape-elk` to make edits directly to the built cytoscape elk built files in my nodes modules while editing this PR, and then once I have something working in this PR, moving the edits upstream to cytoscape elk.

saulshanabrook:
I have added support for showing a loading screen when the layout is running:


https://user-images.githubusercontent.com/1186124/129114686-55025673-865d-4dda-8a25-9f73c4e493ac.mov

saulshanabrook:
Note that Dagre is now not being maintained: https://github.com/dagrejs/dagre/issues/318

saulshanabrook:
I am working on moving all of this to its [own redux slice](https://redux-toolkit.js.org/api/createSlice), in order to more clearly compute the graph and support async layouts.

saulshanabrook:
This is ready for review.

I updated the description with a more comprehensive background and feature set.

You have to link the vega typings, from the PR adding the runtime (vega/vega#3237), to have the typescript compiler work locally (`yarn link "vega-typings"`).

domoritz:
Any idea why the CI is not running @saulshanabrook? Also, can you fix the unused variable lint error?

saulshanabrook:
@domoritz Fixed the lint error. I am not sure! Do you have to approve it to run? I know Github recently added something like that to cut down on github actions abuse.

domoritz:
Thank you so much for working on this. Thank you also for your patience while I am dealing with deadlines and the beginning of the semester at the same time. 

* When clicking on a node, we need some indicator that the graph is filtered now and a button to reset. The state needs to be obvious. Or is that what the "clear" button is supposed to do? It doesn't seem to be working for me.

https://user-images.githubusercontent.com/589034/132349058-707a7312-ea70-469e-8240-6104bce4ef9f.mov

* Minor: when clicking a node, the tooltip stays open for the previously clicked node. 
* The layout isn't great yet. Here is a simple bar chart and the graph is really hard to read. Nodes are too close to each other and not aligned well. Could we align nodes along the same verticals to make the flows easier to follow? Maybe we can even put all renders in the same vertical? There are also unconnected nodes floating somewhere in the middle of the layout making it hard to spot them. 

Screen Shot 2021-09-07 at 08 56 19

* would it make sense to animate node positions between filtering operations so one can follow where nodes go?

@kanitw @arvind @jheer and @jhoffswell, check this out. It's super cool. What suggestions do you have?

package.json
```diff
@@ -25,19 +26,26 @@
     "webpack-dev-server": "^4.0.0"
   },
   "dependencies": {
+    "@reduxjs/toolkit": "1.6.1",
+    "@tippyjs/react": "4.2.5",
```
domoritz, line 30: We already have react-tooltip. Please consolidate into one tooltip library. I don't care which one.

package.json
```diff
@@ -55,14 +63,16 @@
...
     "tslib": "^2.3.1",
     "vega-cli": "5.20.2",
     "vega-datasets": "^2.2.0",
     "vega-embed": "^6.18.2",
     "vega-lite": "5.1.0",
     "vega-schema-url-parser": "^2.2.0",
     "vega-themes": "^2.10.0",
-    "vega-tooltip": "^0.27.0"
+    "vega-tooltip": "^0.27.0",
+    "web-worker": "1.0.0"
```
domoritz, line 75: All dependencies should be `^`

domoritz (changes requested):
Also see easier comments. Thank you for the pull request!

saulshanabrook:
@domoritz Thank you so much for the review! I know it's a big PR and I appreciate all the feedback. I have updated the description of this issue with a list of TODO items drawn from your comments and wanted to respond to a few here as well.  I am also happy for any more code or UX feedback from anyone else, if they have thoughts on how to better display things.

> When clicking on a node, we need some indicator that the graph is filtered now and a button to reset. The state needs to be obvious. Or is that what the "clear" button is supposed to do? It doesn't seem to be working for me.

Thanks for bringing this up, I agree it's quite unintuitive at the moment. A few points about how things work currently:

* The "clear" button simply clears recorded pulses, it should not change the selection, unless a pulse is selected. There seems to be a bug, from your video, where the clear button causes the runtime to be deleted as well.
* The "unselect" button, next to the clear one, will unselect a pulse, if you have one selected.
* To select an element, click on an edge or a node. Hold down shift to select multiple. To unselect, click on the background of the graph. The interaction pattern is similar to a crossfilter in Vega, where the nodes shown is the intersection of the three selections (pulse, element, types)
* This allows you to, for example, select a node in the first pulse, to filter them, then select another pulse, to see all nodes related to that first node. I have used this to find children of the root data node in other pulses. This does mean that sometimes a node is selected, but not visible on the screen, since a pulse is selected.

I am happy for suggestions on how to improve this interaction pattern, I think you are the first person to try this out besides myself, so I understand it might not be the most intuitive. For now, I have added a TODO item to add a "Clear Selected Elements" button that pops up on the graph, or in the sidebar, when you have elements selected, which does the same thing as clicking on the background currently. I have also added TODOs to making the "clear" and "unselect" buttons longer to specify they are with regards to the pulses.

> The layout isn't great yet. Here is a simple bar chart and the graph is really hard to read. Nodes are too close to each other and not aligned well. Could we align nodes along the same verticals to make the flows easier to follow? Maybe we can even put all renders in the same vertical? There are also unconnected nodes floating somewhere in the middle of the layout making it hard to spot them.

Thanks for the suggestions, I will add some padding, try to move unconnected nodes out, and work on aligning nodes. I believe ELK attempts to put them in "layers" so I am not sure why they are not all vertically aligned, I will look into that.

With regards to adding constraints based on node type, like putting all renders on the same vertical, I can try that, especially if we know that there are some patterns which always occur, like having all the renders have only one descendent node, a sieve, like you showed in this example.

domoritz:
Thank you. I look forward to trying the revised version. For the layout, I'm not sure yet what kind of constraints based on patterns make sense. As you try different examples, see whether some patterns emerge that could be useful to exploit.

saulshanabrook:
@domoritz I addressed the feedback you left.

Screen Shot 2021-09-13 at 11 53 27 AM


I made a few small tweaks for the layout, increasing padding, and disabling post contraction (layers should be aligned now).

Previously, I had experimented with enabling ["partitioning"](https://www.eclipse.org/elk/reference/options/org-eclipse-elk-partitioning-partition.html) to influence what layer different types of nodes appeared on. However, this sometimes raised errors, depending on the graph, so I disabled it.

Let me know what you think now

domoritz:
Again, thanks for your work on this. I think we are getting close.

* [ ] Add a button tore-center the chart. It's easy to accidentally zoom out too far. Or alternatively limit the zoom range.
Screen Shot 2021-09-23 at 10 13 07

* [ ] fix white gap under tab
Screen Shot 2021-09-23 at 10 13 48

* [ ] Remove the initial animation when you open the tab. It's distracting and not necessary. 

* [ ] Think about a way to not have tooltips prevent panning. Right now, tooltips for lines appear in the center of a line. I think it would be better to put the tooltip on the link closest to the cursor. Moreover, it would be nice to highlight the line as it's hard to see which line is meant when there are many. Alternatively, just put a label on the line (I would prefer that).

https://user-images.githubusercontent.com/589034/134523801-f98e905b-52ba-42a2-81a0-f9f997fe944f.mov

* [ ] Reduce the darkness of the form outlines

Screen Shot 2021-09-23 at 10 17 24

* [ ] Why are data and signal hidden by default?
* [ ] "Selecting a pulse filters the nodes to those that were updated in that pulse and displays their values on hover" add a `.`. Also, how do you select a pulse?
* [ ] In tooltips, we can at least put the `id` property and value into one row to reduce vertical space in the tooltips
Screen Shot 2021-09-23 at 10 27 44

* [ ] Add a legend for the colors of nodes
* [ ] Is there a way to hide axes/legends (as an option) to make data flow graphs easier to read?
* [ ] Align tooltip for tooltip correctly

Screen Shot 2021-09-23 at 10 34 32

* [ ] I think we can make the visualization easier to use. a) add labels to edges b) add most relevant information to the node (not on hover) such as the scale name c) make black edges have more weight in the layout computation (they should be straight if possible). Below is a comparison with https://observablehq.com/@vega/how-vega-works. 

Screen Shot 2021-09-23 at 10 40 41
Screen Shot 2021-09-23 at 10 40 37

domoritz:
@vega/vg-dev if you have feedback, please add it now as I think this is the last iteration.

saulshanabrook:
Thanks for the review, that all sounds good.

In response to a few points:

>  Why are data and signal hidden by default?

I hid them by default since they are not operators or streams, more like pointers to them. I can enable them by default though if you think that is better. Like in the reference dataflow that you displayed, the signal node does not exist, but the operator nodes are labeled with the signal, if it exists.

>  Add a legend for the colors of nodes

Do you have any suggestions on where to put this legend? I could imagine it being quite large potentially, with all the node types.

>  Is there a way to hide axes/legends (as an option) to make data flow graphs easier to read?

Do you have any suggestions on how to trim this from the graph in a consistent way?

> add most relevant information to the node (not on hover) such as the scale name

I can do this, special casing a few node types. My only concern is that it maybe becomes a little less consistent, since some nodes then wouldn't have their type, but some would.

> make black edges have more weight in the layout computation

That makes sense. I will try to see if this is possible with ELK, but it might not be.

domoritz:
> Do you have any suggestions on where to put this legend? I could imagine it being quite large potentially, with all the node types.

Maybe in the sidebar.

>> add most relevant information to the node (not on hover) such as the scale name
> I can do this, special casing a few node types. My only concern is that it maybe becomes a little less consistent, since some nodes then wouldn't have their type, but some would.

I think we can still show the type name (e.g. scale). With a legend it might also be less of an issue.

> Is there a way to hide axes/legends (as an option) to make data flow graphs easier to read?
Do you have any suggestions on how to trim this from the graph in a consistent way?

No. But maybe @jheer has a suggestion.

saulshanabrook:
I have addressed most of your feedback. What isn't done yet is:

1. Adding legend
2. Changing text of some nodes to show relevant information instead of their types
3. Trying to influence graph layout to prefer keeping bold edges straight


>  Add a button tore-center the chart. It's easy to accidentally zoom out too far. Or alternatively limit the zoom range.

Done, I constrained the zoom so its hard to get lost now.

Max:


Screen Shot 2021-09-26 at 11 30 22 AM


Min:


Screen Shot 2021-09-26 at 11 30 10 AM

>  fix white gap under tab

Fixed by removing grey padding on top of dataflower view panel
Screen Shot 2021-09-26 at 11 26 50 AM

>  Remove the initial animation when you open the tab. It's distracting and not necessary.

Removed this, and also removed animation when no nodes are present in both views.

>  Think about a way to not have tooltips prevent panning. Right now, tooltips for lines appear in the center of a line. I think it would be better to put the tooltip on the link closest to the cursor. Moreover, it would be nice to highlight the line as it's hard to see which line is meant when there are many. Alternatively, just put a label on the line (I would prefer that).

I moved the labels back to the edges, as I had them originally.


Screen Shot 2021-09-26 at 11 26 26 AM

>  Reduce the darkness of the form outlines

Done, made them `--light-gray-color`.


Screen Shot 2021-09-26 at 11 26 09 AM

>  Why are data and signal hidden by default?

I didn't change  these, since they aren't in Jeff's example in the notebook, let me know though if I should.

>  "Selecting a pulse filters the nodes to those that were updated in that pulse and displays their values on hover" add a .. Also, how do you select a pulse?

Updated to: "Clicking on a pulse filters the nodes to those that were updated in that pulse and displays their values on hover."

>  In tooltips, we can at least put the id property and value into one row to reduce vertical space in the tooltips

I changed the tooltips to include the value on the same line to save vertical space:
Screen Shot 2021-09-26 at 11 25 23 AM


>  Align tooltip for tooltip correctly

I believe I fixed it, my css styles for the dataflow popups were interfering with the other popups:

Screen Shot 2021-09-26 at 11 47 09 AM

saulshanabrook:
> I still don't understand what it means to select a pulse. How do I use the "filter by pulse"?

You can click on a row in the table, and that selects a pulse. I should make it more clear somehow what it's doing! Let me know if you have suggestions. 

> Also, I noticed that the cursor switches to a pointer when I hover over the table header. That should not be the case, no?

Yep that shouldn't be the case.

> I think data and signals should be visible by default.

Sounds good, will change.

domoritz:
> You can click on a row in the table, and that selects a pulse. I should make it more clear somehow what it's doing! Let me know if you have suggestions.

Oh, I got confused since my chart doesn't have pulses so the table is empty. You should hide the table when it has no rows.

saulshanabrook:
> Oh, I got confused since my chart doesn't have pulses so the table is empty. You should hide the table when it has no rows.

Huh, I think it should always have one pulse? The initial one? If you select that one, then at least you can see the initial values when you hover over each node.

saulshanabrook:
> Also, I noticed that the cursor switches to a pointer when I hover over the table header. That should not be the case, no?

The current behavior was reversed, so that the header hada pointer and the rows did not. I fixed it so the rows had a pointer, and the header did not.

domoritz:
Thank you. Any idea why the CI doesn't run on your fork?

saulshanabrook:
It looks like this PR isn't run either (#591). I just tried editing the github action config to run on PRs and that seems to make it work now.

Also I believe https://github.com/vega/vega/pull/3237 will need to be released before this passes?

.github/workflows/test.yml
```diff
@@ -1,6 +1,6 @@
 name: Test
 
-on: [push]
+on: [push, pull_request]
```
domoritz, line 3: That will cause multiple test runs for pro from internal branches. Take a look at how we fix this issue in e.g. Vega-lite.

domoritz (changes requested):

.github/workflows/test.yml
saulshanabrook: Thanks, fixed!

saulshanabrook:
Yeah we are now getting a test failure since the upstream vega typings PR isn't merged:

> Error: src/features/dataflow/utils/runtimeToGraph.ts(17,8): error TS2307: Cannot find module 'vega-typings/types/runtime/runtime' or its corresponding type declarations.

domoritz:
Thank you for building out this feature. I'll merge this pull request when we have a typings release.

PR #1043, opened 2021-06-07
gh-349. Update Docker file for embedding to the zorko-platform compose

yzhbankov:


Dockerfile
```diff
@@ -13,26 +13,12 @@ RUN apt-get -y install rsync
...
-
 # Run Yarn
-RUN yarn
-
-# Sets the container executable (ENTRYPOINT) as yarn and the default argument (CMD) as start
-# https://docs.docker.com/engine/reference/builder/#entrypoint
-# https://docs.docker.com/engine/reference/builder/#cmd
-ENTRYPOINT ["yarn"]
-CMD ["start", "--host", "0.0.0.0"]
+RUN yarn install
```
domoritz, line 24: Does this command need to change? Do you need to update the readme as well?

domoritz (changes requested):
I don't know what zorko-platform is. Can you provide more context?

Dockerfile
yzhbankov: Sorry, mistakenly created a pull request for the source project instead of the forked one.

domoritz: Ahh, no problem. Excited to hear that you are using the Vega editor. If you have any changes you want to contribute back, I look forward to your prs.

nesterone:
Hi, @domoritz 

It's an experimental project, to bring all vega blocks like `vega-editor`, `vega-rendering-service`, `vega-scalable` etc under one web server with a some sort of content storage.  It's still in prototype mode (just mock data on UI without connection to DB ), however if you interested we can ping you when we would have something integrated

domoritz:
I'm very interested. Ping me when you have something integrated.

PR #1045, opened 2021-06-20
fixed setImmediate bug

siddhant1:
https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate

I am not sure why there isn't support or why was this removed, anyway added a fix

EDIT : I couldn't find an alternative for setImmediate in browsers except resolving a promise so the loop moves to next tick

cc: @domoritz


siddhant1:
There still are some bugs, don't merge it yet

PR #1147, opened 2022-02-15
Replace json parser with jsonc.

joshpoll:
This PR replaces the JSON parser with JSONC. JSONC is an extension of the JSON spec that includes comments and trailing commas. It's supported by Microsoft and used in Monaco and VSCode.

This should provide a significant quality-of-life improvement for people using the editor.

It's still possible to export and share JSON specs, and this is the default behavior. Users can opt-in to JSONC specs when sharing by clicking the "Preserve whitespace, comments, and trailing commas" button.

### DIFF
--- a/src/utils/monaco.ts
+++ b/src/utils/monaco.ts
@@ -44,7 +44,9 @@ const schemas = [
 
 export default function setupMonaco() {
   Monaco.languages.json.jsonDefaults.setDiagnosticsOptions({
-    allowComments: false,
+    // allowComments: false,
+    comments: 'warning',
+    trailingCommas: 'warning',
     enableSchemaRequest: true,
     schemas,
     validate: true,
--- a/src/utils/monaco.ts
+++ b/src/utils/monaco.ts
@@ -1,4 +1,5 @@
 import stringify from 'json-stringify-pretty-compact';
+import {parse} from 'jsonc-parser';
 import * as Monaco from 'monaco-editor/esm/vs/editor/editor.api';
 import {mergeDeep} from 'vega-lite';
 import addMarkdownProps from './markdownProps';
@@ -73,7 +74,7 @@ export default function setupMonaco() {
       return [
         {
           range: model.getFullModelRange(),
-          text: stringify(JSON.parse(model.getValue())),
+          text: stringify(parse(model.getValue())),
         },
       ];
     },
--- a/src/utils/monaco.ts
+++ b/src/utils/monaco.ts
@@ -45,7 +45,6 @@ const schemas = [
 
 export default function setupMonaco() {
   Monaco.languages.json.jsonDefaults.setDiagnosticsOptions({
-    // allowComments: false,
     comments: 'warning',
     trailingCommas: 'warning',
     enableSchemaRequest: true,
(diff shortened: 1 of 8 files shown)


src/utils/monaco.ts
```diff
@@ -44,7 +45,8 @@ const schemas = [
 
 export default function setupMonaco() {
   Monaco.languages.json.jsonDefaults.setDiagnosticsOptions({
-    allowComments: false,
+    comments: 'warning',
+    trailingCommas: 'warning',
```
domoritz, line 49: Smart. So we don't even have to do any extra work to get warning.

yarn.lock
```diff
@@ -5799,6 +5799,11 @@ json5@^2.1.2, json5@^2.2.0:
   dependencies:
     minimist "^1.2.5"
 
+jsonc-parser@^3.0.0:
```
domoritz, line 5802: Does Monaco depend on some jsonc parser already that we could use?

domoritz (commented):
Neat. It's pretty clean overall. 

@arvind do you have any comments on this pull request?

yarn.lock
joshpoll: It looks like Monaco depends on the same jsonc-parser package: https://github.com/microsoft/monaco-editor/blob/7826dee4ea80e7e4542db6934593e370f2305d97/package.json#L43. I think the jsonc-parser interface is a lot more straightforward than the Monaco interface, but it might be possible to wrap Monaco.

src/utils/monaco.ts
joshpoll: Yeah this way users get a warning in the editor, but it doesn't prevent rendering. I just realized that these warnings don't show up in the logs. Is there way to do that/should we do that?

domoritz: It would be nice. 

There must be a way to get the Monaco warning and show the in the console. 

We could probably replace some of the custom json schema validation if we figure this out.

yarn.lock
domoritz: It's a dev dependency in Monaco. 

I'm asking since I would have expected the dependency to already be in the lock file.

src/utils/monaco.ts
domoritz: Yep. The biggest one is JSON schema validation. And because we use different validators, the error messages are also not very consistent.

domoritz: It would be awesome if you could look into this. If you feel it's too much of a change, we can also do it as a follow up pull request and merge this one first.

joshpoll: I looked into it, and I think this would require hooking into the `onDidChangeMarkers` callback, which would require changing the React wrapper around Monaco or switching to the other Monaco editor wrapper @monaco-editor/react.

So I think it should be a separate PR.

joshpoll:
@arvind I changed the parse import names

src/utils/monaco.ts
domoritz: Okay. I own the monaco component so if you send a pull request, I can review and release. Let's do this in a follow up.

PR #1150, opened 2022-02-18
Fix JSONC parse error handling

joshpoll:
I thought I had implemented error handling correctly for the new parser, but I was mistaken.

H/t to @jonathanzong for bringing this to my attention. **In the current release, syntax errors in manual mode cause the editor to fatally crash.**

This PR fixes error handling by removing it from the schema parsing step and placing it in `parseVega`, `parseVegaLite`, and  `parseConfig` where it actually belongs.

I also removed error handling altogether from schema parsing since the JSONC parser never throws, and all syntax errors will be caught downstream.

### DIFF
--- a/src/reducers/index.ts
+++ b/src/reducers/index.ts
@@ -70,29 +70,16 @@ import {
   SET_TOOLTIP,
 } from './../actions/editor';
 import {LocalLogger} from './../utils/logger';
-import {parse as parseJSONC} from 'jsonc-parser';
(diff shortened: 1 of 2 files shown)


src/reducers/index.ts
```diff
@@ -164,7 +151,7 @@ function parseVega(
   currLogger.level(state.logLevel);
 
   try {
-    const spec = parseJSONC(action.spec);
+    const spec = parseJSONCOrThrow(action.spec);
```
domoritz, line 154: Makes me even more convinced that snake case is better than camel case. Oh well.

PR #1379, opened 2023-11-25
Add id search to dataflow viewer

mcnuttandrew:
Screenshot 2023-11-25 at 4 55 52 PM

When debugging things in the vega data flow graph sometimes it can be tedious to have to manually search for a particular node of interest. This adds a _very simple_ id search which allows the user to search for a particular node

### DIFF
--- a/src/components/input-panel/spec-editor/renderer.tsx
+++ b/src/components/input-panel/spec-editor/renderer.tsx
@@ -198,7 +198,7 @@ class Editor extends React.PureComponent<Props> {
       prevProps.editorRef && prevProps.editorRef.deltaDecorations(prevProps.decorations, []);
     }
 
-    if (this.props.parse) {
+    if (this.editor && this.props.parse) {
       this.editor.focus();
       this.editor.layout();
       this.updateSpec(this.props.value, this.props.configEditorString);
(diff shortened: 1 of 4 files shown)


src/components/input-panel/spec-editor/renderer.tsx
```diff
@@ -198,7 +198,7 @@ class Editor extends React.PureComponent<Props> {
       prevProps.editorRef && prevProps.editorRef.deltaDecorations(prevProps.decorations, []);
     }
 
-    if (this.props.parse) {
+    if (this.editor && this.props.parse) {
```
mcnuttandrew, line 201: Not sure what this is, but i couldn't get the editor to run local without it?

domoritz (changes requested):
Thanks for the feature. Let's make this look a bit nicer. Let's all caps ID. Maybe we can add some spacing to the button, make it inline with the input (IDs are short), make the button full width like the input, or even remove the button and search immediately on change. Also, should we use type search (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search)?

Screenshot 2023-11-26 at 11 37 23

mcnuttandrew:
Thanks for this thoughtful feedback. I believe i've addressed all of it, you can see a preview here:

Screenshot 2023-11-28 at 5 41 34 PM

> or even remove the button and search immediately on change

I found this to be a little bit jerky when i was trying it out, but happy to change over if you feel strongly

Issue #1305, opened 2023-02-10
Not able to run editor locally following instructions in Readme


ArulselvanMadhavan:
Hello, Thanks for maintaining this library!

I have not been able to run the editor in my machine following the instructions in the Readme. At first, the yarn start throws an error saying "tsconfig.json" is missing. I noticed that it was added to gitignore. I found an older version in the commit history and bypassed that error. Then, `yarn start` failed stating that public/spec/vega/index.json and vega-lite/index.json are missing. I found older versions from the commit history to start the server.

Could you please update the Readme with instructions on how to run the editor locally?

domoritz:
if you run `yarn`, it should set up your environment including the tsconfig. You can see the build steps in https://github.com/vega/editor/blob/master/.github/workflows/test.yml. 

Can you post your environment details?

schaetzc:
This issue is older than a year, not sure if my problem was the same one:

I cloned the repo (currently on commit 691d6195b75ad34f7afa8dc526f3a5561391fd37) and ran `yarn`. This resulted in ...
```
yarn install v1.22.19
info No lockfile found.
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[4/4] Building fresh packages...

success Saved lockfile.
Done in 0.07s.
root@1deaaa1511e4:/usr/app# cd editor/
root@1deaaa1511e4:/usr/app/editor# yarn
yarn install v1.22.19
[1/4] Resolving packages...
[2/4] Fetching packages...
[####------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 16/812^C
root@1deaaa1511e4:/usr/app/editor# time yarn
yarn install v1.22.19
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
warning " > react-split-pane@0.1.92" has incorrect peer dependency "react@^16.0.0-0".
warning " > react-split-pane@0.1.92" has incorrect peer dependency "react-dom@^16.0.0-0".
warning " > vega-lite@5.20.1" has unmet peer dependency "vega@^5.24.0".
warning " > vega-embed@6.26.0" has unmet peer dependency "vega@^5.21.0".
warning " > vega-themes@2.15.0" has unmet peer dependency "vega@*".
warning " > autoprefixer@10.4.20" has unmet peer dependency "postcss@^8.1.0".
[4/4] Building fresh packages...
$ husky && npx patch-package
[...]

PR #1444, opened 2024-09-15
Better usage instructions and less build requirements (rm+cp instead of rsync)

schaetzc:
This pull request improves documentation and build scripts for users who just want to run vega-editor locally.
See https://github.com/vega/editor/issues/1305 for more information.

`rsync` is no longer needed to run install/build/run vega-editor. Using standard `rm` and `cp` commands instead of `rsync` is not only more portable, but was also faster on my system. I ran multiple repeated tests ...
- on an SSD and on a 10 year old HDD,
- with and without dropping caches (`sudo sh -c 'sync && echo 3 > /proc/sys/vm/drop_caches'`) in between repeats,
- for situations where the target directories were only updated, and situations where they were initialized for the first time.

Sometimes `rsync` took 10 times as long. But it does not really matter. Copying always took < 0.5 s while the downloads from the internet were the main slowdown for me.

This pull request also fixes some bugs in `scripts/vendor.sh`, e.g.
- Spaces and special symbols ``* ? [] $ ` `` are now supported in the (absolute) path of all files involved.
- When an example is removed from the source dirs, it is now removed in the (existing) target dirs too (basically `rsync`'s `--delete` option, which wasn't used so far)

Documentation was improved, so that other requirements (which are not changed in this pull request) are explicitly mentioned.

The oddity that the working copy cannot be directly below the root directory comes from below line in `package.json` ...
```json
"scripts": {
  "start": "PARCEL_BUILD_COMMIT_HASH=$(git rev-parse HEAD) parcel --watch-dir .. index.html",
```
The parent directory `..` could be `/` which `parcel` does not like. In such a situation it prints `[Error: Invalid argument]` and terminates with exit code 1.  
This could be fixed by changing `parcel --watch-dir ..` to `parcel --watch-dir .` or leaving out the option completely. However, I am not sure if that would have negative effects on the developers. I assume they have multiple vega subprojects checked out in one shared top level directory `vega/` (e.g. `vega/editor`, `vega/lite`, ...) and actually want the `..` . Therefore I only commented on it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ You can reset the Vega Editor by going to https://vega.github.io/editor/#/reset 
 
 ## Usage Instructions
 
-To run the editor locally, you must first install the dependencies and then launch a local web server. We assume you have [yarn](https://yarnpkg.com/) installed.
+To run the editor locally, you must first install the dependencies and then launch a local web server. We assume you have [yarn](https://yarnpkg.com/), `bash`, `curl`, and `tar` installed.
 
 1. Install the dependencies:
 
--- a/README.md
+++ b/README.md
@@ -10,21 +10,26 @@ You can reset the Vega Editor by going to https://vega.github.io/editor/#/reset 
 
 ## Usage Instructions
 
-To run the editor locally, you must first install the dependencies and then launch a local web server. We assume you have [yarn](https://yarnpkg.com/), `bash`, `curl`, and `tar` installed.
-
-1. Install the dependencies:
-
-```
-$ yarn
+### Run Vega-Editor With Docker
+```bash
+sudo docker run -it --rm -p 1234:1234 node:21 bash -xc 'mkdir -p vega/editor && git clone --depth=1 https://github.com/vega/editor.git vega/editor && cd vega/editor && yarn && yarn start'
 ```
 
-2. Start the server:
+### Development Setup
+We assume you have [yarn](https://yarnpkg.com/), `bash`, `curl`, and `tar` installed.
 
-```
-$ yarn start
-```
+Your working copy of this git repository must be located at least two levels below the system root `/`.
+E.g. `/home/user/editor` or `/vega/editor`, but not `/editor`. 
 
-3. The local web server will be accessible from [http://localhost:8080](http://localhost:8080).
+Inside your working copy ...
+
+1. Install the dependencies:  
+   `$ yarn`
+
+2. Launch the local web server:  
+   `$ yarn start`
+
+3. The local web server will be accessible via [http://localhost:1234](http://localhost:1234).
 
 ## Local Testing & Debugging
 
--- a/scripts/vendor.sh
+++ b/scripts/vendor.sh
@@ -4,37 +4,28 @@ set -e
 
 DATA=public/data
 SPEC=public/spec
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -10,21 +10,29 @@ You can reset the Vega Editor by going to https://vega.github.io/editor/#/reset
 
 ## Usage Instructions
 
-To run the editor locally, you must first install the dependencies and then launch a local web server. We assume you have [yarn](https://yarnpkg.com/) installed.
+### In Short
```
domoritz, line 13: Maybe rename this to "running with docker"

README.md
```diff
@@ -10,21 +10,29 @@ You can reset the Vega Editor by going to https://vega.github.io/editor/#/reset
...
+```bash
+sudo docker run -it --rm -p 1234:1234 node:21 bash -xc 'mkdir -p vega/editor && git clone --depth=1 https://github.com/vega/editor.git vega/editor && cd vega/editor && yarn && yarn start'
+```
 
-1. Install the dependencies:
+### In Detail
```
domoritz, line 18: And this to "development setup"

README.md
```diff
@@ -10,21 +10,29 @@ You can reset the Vega Editor by going to https://vega.github.io/editor/#/reset
...
+We assume you have the following tools installed:
+- [yarn](https://yarnpkg.com/)
+- The older `node` version 21, because of an incompatibility in one of our dependencies, see [node-canvas issue](https://github.com/Automattic/node-canvas/issues/2377)
```
domoritz, line 21: This feels out of place in the readme since we can hopefully fix this going forward and will forget to update it here. What if we set the node version in the package.json instead?

domoritz (approved):
Thank you for also fixing other things. Just two comments and otherwise good to go.

schaetzc:
Thank you for the review.

I `--amend`ed my last git commit, to adapt the subsection titles as suggested, and remove the overly specific detail regarding the node version.  
I assume pressing the "*resolve conversation*" buttons in this pull request is up to you, right?

README.md
schaetzc: Updated to "_Run Vega-Editor With Docker_"

README.md
schaetzc: Updated to "_Development Setup_"

README.md
schaetzc: Yeah, I thought the same too, but wasn't sure where to put it instead.

> What if we set the node version in the package.json instead?

If that is possible and an acceptable practice, that would be great. I have no experience with tools like node and yarn, and cannot judge that.  
I assumed there was a reason node wasn't already managed through package.json or yarn.lock, e.g. that the tools installing/updating the packages cannot modify the node installation while they run inside it.

I removed the hint without changing anything else. I'd be happy if you could handle that. But I'm also ok with merging as is. By reading the docker command, the information which node version to use is still there.

> hopefully fix this going forward

I was hoping so too. It seems there will be no release for node-canvas 2.x anymore; they already have a release preview for 3.0. So to make use of the upstream fix, vega-editor  probably has to upgrade to node-canvas 3.x once its officially released. I have no idea, if there are any breaking changes that could affect vega-editor.

domoritz:
Thank you

PR #1454, opened 2024-11-27
README.md development setup for apple silicon and others

MarcBaeuerle:
- [x] This PR is atomic (i.e., it fixes one issue at a time).
- [x] The title is a concise [semantic commit message](https://www.conventionalcommits.org/) (e.g. "fix: correctly handle undefined properties").

I noticed a week ago you migrated from `yarn` to `npm`, however I tried setting up the project for the first time today and I was running into issues when running `npm install` (`command failed: yarn run postinstall-postinstall`), until I used the correct yarn version (v1). Originally included this in the readme but removed since I don't know if that will be fixed, should I create an issue?

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,9 +25,7 @@ Inside your working copy ...
 
 1. Install the dependencies:  
    `$ npm install`
-   * macOS Apple Silicon: if you encounter installation errors related to `canvas`, ensure the necessary system libraries are present by running:   
-     `$ brew install pkg-config cairo pango libpng jpeg giflib librsvg pixman python-setuptools`  
-     For more details, refer to the [canvas installation guide](https://github.com/Automattic/node-canvas?tab=readme-ov-file#installation)
+   * If you are running into issues with installing canvas, follow the [canvas installation guide](https://github.com/Automattic/node-canvas?tab=readme-ov-file#installation).
 
 2. Launch the local web server:  
    `$ npm start`


domoritz:
Why is this arm specific or depends on the node version?

Doesn't the error message already describe what packages may be needed?

MarcBaeuerle:
arm specific: the error messages just stated that it failed to fetch canvas packages, as there are no prebuilt binaries that can be downloaded for apple silicon, so the workaround is needed. No clear error messages and StackOverflow threads were outdated.

yarn: npm install only worked with yarn v1, and failed with yarn v2 v3 and v4. There were no error messages that indicated you need to switch versions, I only knew this because I faced a similar issue before.

domoritz:
Hmm, which node version do you have. I just tried in my M1 Mac and didn't need to build canvas with node 20.

```
~/Developer/editor on  master via  v20.18.0 took 13s ❯ trash node_modules/
~/Developer/editor on  master via  v20.18.0 ❯ npm i
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated npmlog@5.0.1: This package is no longer supported.
npm warn deprecated rimraf@2.7.1: Rimraf versions prior to v4 are no longer supported
npm warn deprecated @humanwhocodes/config-array@0.13.0: Use @eslint/config-array instead
npm warn deprecated rimraf@3.0.2: Rimraf versions prior to v4 are no longer supported
npm warn deprecated are-we-there-yet@2.0.0: This package is no longer supported.
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
npm warn deprecated @humanwhocodes/object-schema@2.0.3: Use @eslint/object-schema instead
npm warn deprecated gauge@3.0.2: This package is no longer supported.
npm warn deprecated eslint@8.57.1: This version is no longer supported. Please see https://eslint.org/version-support for other options.

> postinstall
> husky && npx patch-package
...
```

README.md
```diff
@@ -25,6 +25,9 @@ Inside your working copy ...
 
 1. Install the dependencies:  
    `$ npm install`
+   * macOS Apple Silicon: if you encounter installation errors related to `canvas`, ensure the necessary system libraries are present by running:   
+     `$ brew install pkg-config cairo pango libpng jpeg giflib librsvg pixman python-setuptools`  
+     For more details, refer to the [canvas installation guide](https://github.com/Automattic/node-canvas?tab=readme-ov-file#installation)
```
domoritz, line 30:
```suggestion
   * If you are running into issues with installing canvas, follow the [canvas installation guide](https://github.com/Automattic/node-canvas?tab=readme-ov-file#installation).
```

PR #1461, opened 2025-01-26
feat(editor): Add setting for using editor with `expressionInterpreter` enabled

### LINKED ISSUE #1460: Feature: Allow running graph preview with `expressionInterpreter` mode enabled
## Motivation

- When investigating reports into behaviors of custom expressions like https://github.com/vega/vega/issues/3984, it would be beneficial to let maintainers rapidly test running `vega` with the [expression interpreter](https://vega.github.io/vega/usage/interpreter/) (AST) mode enabled. 
- This allows testing extent of the bug without having to set up a local dev environment

## Proposal

- Add a checkbox to the settings bar (similar to the tooltip checkbox) which allows toggling this behavior

Out of scope (for now):
-  Do not persist this state to URL parameters (as most of the other app states are not URL sync'd either, aside from the VL spec itself). That could be investigated in a follow-up.

hydrosquall:
## Motivation

- Faster testing of vegaExpression related bug reports
- Fixes #1460 

## Changes

| Before | After |
| --- | --- |
| image|  image |

- Adds `expressionInterpreter` to the root redux state. (Open to naming this so it looks more like a boolean, but since the name is long already I figured this was OK to start), and wires up associated Redux boilerplate (reducers, action creators, etc)
- Adds a settings toggle to the sidepanel (pictured)
- When I saved the files, some extra formatting was applied by my VSCode defaults. At first I was going to revert the changes, but then I saw the styles match what  you get when running `yarn format $(modified_files)`, so I left them as is.

## Testing

- This will have to be tested locally, since we haven't set up branch previews yet ( https://github.com/vega/vega-lite/issues/9276 ) . We could have a test case where we know the behavior of the `expressionInterpreter` is slightly different.


src/components/renderer/renderer.tsx
```diff
@@ -126,15 +127,27 @@ class Editor extends React.PureComponent<Props, State> {
...
 
     let runtime: vega.Runtime;
     if (mode === Mode.VegaLite) {
       // In vl mode, we compile Vega-Lite spec along with config to Vega spec
-      runtime = vega.parse(vegaSpec);
+      runtime = vega.parse(vegaSpec, {}, parseOptions);
     } else {
-      runtime = vega.parse(vegaSpec, config as VgConfig);
+      runtime = vega.parse(vegaSpec, config as VgConfig, parseOptions);
     }
```
domoritz, line 151: Could probably simplify this to `mode === Mode.VegaLite ? {} : config`

domoritz (approved):
Thank you. Looks great.

src/components/renderer/renderer.tsx
hydrosquall: Good idea, just pushed the change 👍

hydrosquall:
CI not passing seems to be unrelated to this diff (I opened a separate issue for it https://github.com/vega/editor/issues/1462 ). 

Would you recommend merging the new setting first, or doing the CI fix first?

domoritz:
I'm okay with either. Having a ci fix first would be nicer since then we can merge this one cleanly.

hydrosquall:
Fixed CI in https://github.com/vega/editor/pull/1463, merging this since the checks are green.

PR #1488, opened 2025-05-15
fix(auth): update auth to use localStorage

yuvvantalreja:
## Problem
The application was experiencing authentication issues specifically in Safari:
1. Cookies weren't being properly sent with requests in Safari due to its stricter third-party cookie policies
2. When checking authentication status, Safari would return 304 Network Error
3. Logout functionality wasn't working in Safari

## Solution

Implemented a token-based authentication system as a workaround:

1. Added localStorage-based authentication token storage:
   - Created browser utility functions to manage auth data in localStorage
   - Implemented token generation and validation on the backend

2. Modified frontend authentication flow:
   - Updated components to use localStorage auth data when needed
   - Added header token authentication for API requests

### DIFF
--- a/src/components/gist-select-widget/renderer.tsx
+++ b/src/components/gist-select-widget/renderer.tsx
@@ -4,6 +4,7 @@ import ReactPaginate from 'react-paginate';
 import {mapDispatchToProps, mapStateToProps} from './index.js';
 import {BACKEND_URL, GistPrivacy} from '../../constants/index.js';
 import LoginConditional from '../login-conditional/index.js';
+import {isSafari} from '../../utils/browser.js';
 import './index.css';
 
 interface State {
@@ -80,11 +81,25 @@ class GistSelectWidget extends React.PureComponent<Props, State> {
           loading: false,
         },
         async () => {
+          const headers: HeadersInit = {
+            'Cache-Control': 'no-cache, no-store, must-revalidate',
+            Pragma: 'no-cache',
+            Expires: '0',
+          };
+
+          if (isSafari()) {
+            const token = localStorage.getItem('vega_editor_auth_token');
+            if (token) {
+              headers['X-Auth-Token'] = token;
+            }
+          }
+
           let response;
           if (page.selected === 0) {
             response = await fetch(`${BACKEND_URL}gists/user?cursor=init&privacy=${this.props.private}`, {
               credentials: 'include',
               method: 'get',
+              headers,
             });
           } else {
             response = await fetch(
@@ -92,6 +107,7 @@ class GistSelectWidget extends React.PureComponent<Props, State> {
               {
                 credentials: 'include',
                 method: 'get',
+                headers,
               },
             );
           }
--- a/src/components/header/renderer.tsx
+++ b/src/components/header/renderer.tsx
@@ -69,25 +69,30 @@ class Header extends React.PureComponent<PropsType, State> {
     try {
       const response = await fetch(`${BACKEND_URL}auth/github/check`, {
         credentials: 'include',
+        cache: 'no-store',
       });
       const data = await response.json();
       const {isAuthenticated, handle, name, profilePicUrl} = data;
       this.props.receiveCurrentUser(isAuthenticated, handle, name, profilePicUrl);
     } catch (error) {
-      console.error(error);
+      console.error('Initial authentication check failed:', error);
+      this.props.receiveCurrentUser(false, '', '', '');
     }
 
     window.addEventListener('message', async (e) => {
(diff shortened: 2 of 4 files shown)


src/components/header/renderer.tsx
```diff
@@ -66,28 +72,128 @@ class Header extends React.PureComponent<PropsType, State> {
       }
     });
 
+    const usingLocalStorage = isSafari();
```
domoritz, line 75: Can we always use local storage, not just for safari?

yuvvantalreja: Yes, absolutely. I'll do that.

yuvvantalreja:
## Updates:
- Unified auth to use local storage on all browsers (bdc678ff8f0ded8eb24f53f5f42b5cb4ca7ad853)
- Implemented new architecture that allows direct GitHub API access from the frontend (866d5dc9980e5154f651f8b6a3bb3ded6de12dba)

Note: Still getting a build issue while the lint and build workflow is running on the PR

src/components/gist-select-widget/renderer.tsx
```diff
@@ -80,37 +84,82 @@ class GistSelectWidget extends React.PureComponent<Props, State> {
...
-                method: 'get',
+          try {
+            const authData = getAuthFromLocalStorage();
+            const githubToken = authData?.githubAccessToken || localStorage.getItem('vega_editor_github_token');
+
+            if (!githubToken) {
+              this.props.receiveCurrentUser(false);
+              return;
+            }
+            const privacy = this.props.private ? 'all' : 'public';
```
domoritz, line 95: This is unused

src/types.d.ts
```diff
@@ -61,3 +61,186 @@ declare module 'react-router-dom' {
     staticContext?: any;
   } & T;
 }
+
+declare module 'react-split-pane-r17' {
```
domoritz, line 65: can we not get this from a package? Feels odd to have so many types here manually that are not from the authors.

yuvvantalreja: I couldn't find an official type package for clipboard and observer, so I thought to write them manually for consistency.

I can definitely use packages for the others.

yuvvantalreja: I have tried using packages as much as possible, but for a lot of the components, our usage isn't compatible with the default type definitions. I think this is because of version discrepancy between the official supported type defs and our implementation. Do let me know if I can approach this differently.

domoritz: We could match the versions, replace the dependency, or just not have types. I'd say let's deal with this in a separate pull request.

src/components/app.tsx
```diff
@@ -120,14 +121,58 @@ class App extends React.PureComponent<PropsType> {
 
   public async setGist(parameter: {id: string; filename: string; revision?: string}) {
     try {
-      const gistResponse = await fetch(
-        `https://api.github.com/gists/${parameter.id}${
-          parameter.revision !== undefined ? `/${parameter.revision}` : ''
-        }`,
-      );
+      const gistApiUrl = `https://api.github.com/gists/${parameter.id}${
```
domoritz, line 124: This was weird now that we don't proxy gist requests anymore. I changed the url in the server.

src/components/header/renderer.tsx
```diff
@@ -170,6 +212,7 @@ class Header extends React.PureComponent<PropsType, State> {
           }
 
           const response = await fetch(`${BACKEND_URL}auth/github/check`, {
+            headers,
             credentials: 'include',
```
domoritz, line 216: We should probably delete this here and all other places, no?

src/components/app.tsx
domoritz: Why do we still need this actually?

yuvvantalreja: Doesn't the frontend still have to send requests to load the visualisations from github?

src/components/header/renderer.tsx
```diff
@@ -66,28 +67,172 @@ class Header extends React.PureComponent<PropsType, State> {
...
+
+        const tokenData = await this.verifyTokenLocally(tokenFromHash);
+        if (tokenData && tokenData.isAuthenticated) {
+          saveAuthToLocalStorage({
+            isAuthenticated: tokenData.isAuthenticated,
+            handle: tokenData.handle,
+            name: tokenData.name,
+            profilePicUrl: tokenData.profilePicUrl,
+            authToken: tokenData.authToken,
+          });
```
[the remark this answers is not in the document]
domoritz: We also don't store the GitHub token in local storage exactly because of xss concerns.

PR #1489, opened 2025-05-19
TS type fix

yuvvantalreja:
### Problem

Was experiencing TypeScript errors during the GitHub workflow's npx tsc step. These errors were primarily related to missing or incompatible type declarations for several external libraries, causing JSX element type errors and constructor signature issues.

### Proposed Solution

- Installing official type declarations where available
- Maintaining custom type declarations for libraries don't have official TypeScript definitions, have official types that are incompatible with our current usage patterns or, are forked versions of libraries.

However, this is not a scalable solution for long-term use.

### DIFF
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -61,3 +61,186 @@ declare module 'react-router-dom' {
     staticContext?: any;
   } & T;
 }
+
+declare module 'react-split-pane-r17' {
+  import React from 'react';
+
+  interface SplitPaneProps {
+    split?: 'vertical' | 'horizontal';
+    minSize?: number;
+    maxSize?: number;
+    defaultSize?: number | string;
+    size?: number | string;
+    step?: number;
+    onDragStarted?: () => void;
+    onDragFinished?: (newSize: number) => void;
+    onChange?: (newSize: number) => void;
+    pane1Style?: React.CSSProperties;
+    pane2Style?: React.CSSProperties;
+    paneStyle?: React.CSSProperties;
+    className?: string;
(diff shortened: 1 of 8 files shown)


yuvvantalreja:
> ### Problem
> Was experiencing TypeScript errors during the GitHub workflow's npx tsc step. These errors were primarily related to missing or incompatible type declarations for several external libraries, causing JSX element type errors and constructor signature issues. ([42cdc98](https://github.com/vega/editor/commit/42cdc98eb3569f15c31950765b581e70002433f7))
> 
> ### Proposed Solution
> * Installing official type declarations where available
> * Maintaining custom type declarations for libraries don't have official TypeScript definitions, have official types that are incompatible with our current usage patterns or, are forked versions of libraries.
> 
> However, this is not a scalable solution for long-term use.

I've just removed types for these components altogether (e924abd147e7b9c7b7ecf53dc1325b611e33cfe5)

Please let me know if this is an appropriate solution

src/types.d.ts
```diff
@@ -1,5 +1,9 @@
 /* eslint-disable @typescript-eslint/ban-types */
 
+// TODO: Use default types from packages
+// Current Issues: For a lot of the components,
+// our usage isn't compatible with the default type definitions.
+
```
yuvvantalreja, line 6: addressed in (36e408ebae50563d1c3125396b6248b05b8d32e9)

domoritz:
I am pretty against maintaining our own types. I'd rather have `any` in places. Writing our own types is a maintenance burden. I am worried that we don't keep up with changes in the code.

PR #1494, opened 2025-06-13
Fix/react upgrade

yuvvantalreja:


domoritz:
The resize handling seems to be somewhat broken on safari. Can you fix that? 

https://github.com/user-attachments/assets/9f653025-20d7-494a-b19a-171653e0fe33

domoritz:
I converted this to a draft. When you flip the switch, I can review again.

domoritz:
Strange issue. When I first open the table view, there is an error. It shows up the second time. 
Screenshot 2025-07-24 at 12 38 53
Screenshot 2025-07-24 at 12 38 56

yuvvantalreja:
> The resize handling is much better. The color (#ebe6e6) seems to be a new color, though. can we reuse an existing color? Also, is there a way to make the target are bigger? (not the visual one but the area you can drag with).

I was making the resize component styling consistent with IDEs like VSCode. But sure I can reuse the existing design.

PR #1495, opened 2025-06-15
Transition to vite

yuvvantalreja:
Transition editor from parcel to vite

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     ]
   },
   "scripts": {
-    "start": "vite --port 1234",
+    "start": "vite",
     "build": "npm run clean && npm run vendor && npm run build:only",
     "build:only": "vite build",
     "preview": "vite preview",
--- a/index.html
+++ b/index.html
@@ -20,7 +20,7 @@
 <body>
   <div id="root"></div>
   <noscript>You need to enable JavaScript to run this app.</noscript>
-  <script type="module" src="/src/index.tsx"></script>
+  <script type="module" src="./src/index.tsx"></script>
 </body>
 
 </html>


vite.config.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export default defineConfig({
+  plugins: [
+    react(),
+    legacy({
+      targets: ['defaults', 'not IE 11'],
+    }),
+  ],
+  define: {
+    'process.env.PARCEL_BUILD_COMMIT_HASH': JSON.stringify(commitHash),
```
domoritz, line 16: Oh, I see. That's weird. Can we just have one name?

domoritz:
Let's check whether HMR or at least updating the front end works when linked dependencies for Vega or Vega-Lite change.

vite.config.ts
```diff
@@ -4,29 +4,59 @@ import legacy from '@vitejs/plugin-legacy';
...
+      include: '**/*.{jsx,tsx,ts,js}',
     }),
+    linkedDependencyHMR(),
   ],
   define: {
     'process.env.PARCEL_BUILD_COMMIT_HASH': JSON.stringify(commitHash),
     'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
   },
-  base: '/editor/',
+  base: process.env.NODE_ENV === 'production' ? '/editor/' : '/',
```
domoritz, line 36: why do we need this?

vite.config.ts
```diff
@@ -4,29 +4,59 @@ import legacy from '@vitejs/plugin-legacy';
 import {resolve} from 'path';
 
 const commitHash = process.env.VITE_COMMIT_HASH;
+function linkedDependencyHMR() {
+  return {
+    name: 'linked-dependency-hmr',
+    handleHotUpdate({file, server}) {
+      if (file.includes('node_modules/vega-lite/src/') || file.includes('node_modules/vega-lite/build/')) {
```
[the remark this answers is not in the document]
yuvvantalreja: Sure, will update it!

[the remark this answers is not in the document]
yuvvantalreja: Updated to support all vega packages

package.json
```diff
@@ -12,6 +12,7 @@
     "@types/node": "^22.13.14",
     "@typescript-eslint/eslint-plugin": "^7.15.0",
     "@typescript-eslint/parser": "^7.15.0",
+    "@vitejs/plugin-legacy": "^5.2.0",
```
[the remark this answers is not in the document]
yuvvantalreja: Sure, I'll remove it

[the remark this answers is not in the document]
domoritz: Great. In general, let's remove dependencies when possible. Leaner is better.

domoritz:
Thanks for the quick changes. I am getting an error when I use a linked Vega-Lite. Do you see the same?

Screenshot 2025-06-17 at 16 27 22

domoritz:
Also heads up for @hydrosquall that we may need to change the ci builds after these changes. Hopefully it makes things easier.

yuvvantalreja:
> Thanks for the quick changes. I am getting an error when I use a linked Vega-Lite. Do you see the same?
> 
> Screenshot 2025-06-17 at 16 27 22

Did you build vega-lite before linking?

vite.config.ts
```diff
@@ -0,0 +1,84 @@
...
+        '!**/node_modules/vega/src/**',
+        '!**/node_modules/vega/build/**',
+      ],
+    },
+    fs: {
+      allow: ['..'],
+    },
+  },
+  optimizeDeps: {
+    exclude: ['vega-lite', 'vega'],
```
domoritz, line 80: needs the Vega packages

vite.config.ts
```diff
@@ -0,0 +1,84 @@
...
+  return {
+    name: 'linked-dependency-hmr',
+    handleHotUpdate({file, server}) {
+      if (file.includes('node_modules/vega-lite/src/') || file.includes('node_modules/vega-lite/build/')) {
+        const module = server.moduleGraph.getModuleById('vega-lite');
+        if (module) {
+          server.reloadModule(module);
+        }
+        server.ws.send({
+          type: 'full-reload',
```
domoritz, line 16: This isn't doing full HMR, right? We just happen to have the state in local storage.

vite.config.ts
```diff
@@ -0,0 +1,84 @@
...
+        });
+        return [];
+      }
+
+      if (file.includes('node_modules/vega')) {
+        const regex = /node_modules\/(vega[^/]*)/;
+        const match = file.match(regex);
+
+        if (match && match[1]) {
```
domoritz, line 25: this match doesn't really exclude things. probably remove it

domoritz (changes requested):
Look at https://github.com/vega/editor/pull/1420/files to see what else we might undo

tsconfig.json
```diff
@@ -10,6 +10,13 @@
     "importHelpers": true,
     "jsx": "react",
     "resolveJsonModule": true,
+    "baseUrl": ".",
+    "paths": {
+      "vega-lite": ["../../vega-lite/src"],
+      "vega-lite/*": ["../../vega-lite/src/*"],
```
domoritz, line 16: does that actually work since we only allow some exports defined in https://github.com/vega/vega-lite/blob/473f3594e3db5c24f439f0c183eac8f662f46817/package.json#L27 ?

vite.config.ts
```diff
@@ -1,66 +1,160 @@
...
-              if (module.id && (module.id.includes('vega-lite') || module.url?.includes('vega-lite'))) {
-                vegaLiteModules.push(module);
-                moduleGraph.invalidateModule(module);
+        const nodeModulesPath = resolve(__dirname, 'node_modules');
+        const vegaPackagePaths: string[] = [];
+        const vegaPackageAliases: Record<string, string> = {};
+
+        const items = readdirSync(nodeModulesPath);
+        for (const item of items) {
+          if (item === 'vega' || item.startsWith('vega-')) {
```
domoritz, line 34: duplicated logic?

yuvvantalreja:
I've made the vite config as maintainable as possible. Removed the duplicate logic, simplified conditions and eliminated unnecessary lines.

package.json
```diff
@@ -99,9 +98,10 @@
     ]
   },
   "scripts": {
-    "start": "PARCEL_BUILD_COMMIT_HASH=$(git rev-parse HEAD) parcel --watch-dir .. index.html",
+    "start": "VITE_COMMIT_HASH=$(git rev-parse HEAD) vite --port 1234",
```
domoritz, line 101: do we need that? It wasn't before https://github.com/vega/editor/pull/1420/files

src/constants/specs.ts
```diff
@@ -1,4 +1,23 @@
-import VEGA_SPECS from '../../public/spec/vega/index.json';
-import VEGA_LITE_SPECS from '../../public/spec/vega-lite/index.json';
+import {getPublicPath} from '../utils/publicPath';
```
domoritz, line 1: Why is this needed? Instead can we load just like before in https://github.com/vega/editor/pull/1420/files?

src/utils/logger.ts
```diff
@@ -1,5 +1,5 @@
-import {store} from './../index.js';
```
domoritz, line 1: so do we still need `'./../index.js';`? Seems odd to use a different store in different places of the code.

src/utils/logger.ts
```diff
@@ -47,23 +47,40 @@ export class LocalLogger implements vega.LoggerInterface {
...
 
 export class DispatchingLogger implements vega.LoggerInterface {
+  private store: Store | null = null;
+
+  public initializeStore(store: Store) {
+    this.store = store;
+  }
+
   public level = (_?: number) => {
+    if (!this.store) {
```
domoritz, line 57: when does this happen?

src/utils/logger.ts
```diff
@@ -86,10 +109,16 @@ export class DispatchingLogger implements vega.LoggerInterface {
   };
 
   public debug = (...args: any[]) => {
+    if (!this.store) {
```
domoritz, line 112: when does this happen?

src/utils/publicPath.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Helper function to resolve public directory assets
+ * @param path Path relative to public directory
+ * @returns Full URL to the asset
+ */
+export function getPublicPath(path: string): string {
```
domoritz, line 6: This feels icky. Vite is supposed to handle this stuff.

domoritz (changes requested):

package.json
yuvvantalreja: I was trying to keep the naming consistent

src/utils/logger.ts
yuvvantalreja: I added this incase of race conditions

domoritz:
Let's use del-cli. Not rm.

package.json
yuvvantalreja: I can use vite serve, but the commit hash is being used in the vizpane and help modal, removing it will make those components break

yuvvantalreja: Oh I see, you're right!

src/utils/logger.ts
domoritz: Thoughts on this @yuvvantalreja ?

yuvvantalreja: Yes, we're no longer using this. Now, the store is common throughout the project:
`import {Store} from 'redux';`

PR #1518, opened 2025-09-29
chore(deps): Upgrade Editor Dependencies

yuvvantalreja:
## Changes:

- Upgraded outdated dependencies
- Removed unnecessary deps (husky, patch, url, assert)


domoritz (approved):

package.json
```diff
@@ -9,83 +9,73 @@
...
     "autoprefixer": "^10.4.21",
-    "browserify-zlib": "^0.2.0",
-    "del-cli": "^6.0.0",
-    "eslint": "^8.57.0",
-    "eslint-config-prettier": "^9.1.0",
-    "eslint-plugin-prettier": "^5.5.3",
-    "eslint-plugin-react": "^7.37.4",
-    "events": "^3.3.0",
+    "del-cli": "^7.0.0",
+    "eslint": "^9.36.0",
```
domoritz, line 22: Upgrading aslant requires a new config file. You could use the one from https://github.com/vega/vega-embed/blob/main/eslint.config.js and remove rules that are not relevant/add other relevant rules.

domoritz:
I think you can get browser globals from the globals package instead of listing them manually.

domoritz (approved):

